PR #2, opened 2019-10-22
Added config options to enable HTTPS and custom vhost names

onitake:
Moved from https://github.com/shapeblue/primate-old-do-not-use/pull/41 .

This PR is a proposed solution for https://github.com/shapeblue/primate-old-do-not-use/issues/37 .

Not everybody may want to configure HTTPS, it's probably necessary to make some of the env vars optional.

### DIFF
--- a/.env.local.example
+++ b/.env.local.example
@@ -1,1 +1,7 @@
 CS_URL=http://localhost:8080
+PUBLIC_HOST=primate.example.com
+HTTPS_CERT=/etc/ssl/certs/primate.example.com.pem
+HTTPS_KEY=/etc/ssl/private/primate.example.com.key
+HTTPS_CA=/etc/ssl/certs/ca.pem
+HTTPS_DHPARAM=/etc/ssl/keys/dh2048.pem
+ALLOWED_HOSTS=["primate.example.com","cloud.example.com"]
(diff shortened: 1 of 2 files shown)


rhtyd:
@onitake Would this break the default development work, if someone tries this without having any of the mentioned certificates?

onitake:
Yes, it would indeed.
The problem is that I'm not familiar with JavaScript and vue.js, so I don't know what the best way to make these settings conditional would be.
Something like
```javascript
if (env.is_set('HTTPS_CERT') {
    devserver['https']['cert'] = fs.readFileSync(process.env.HTTPS_CERT)
}
```
maybe?
I'm sure there's a more elegant way to do it in JS.

onitake:
Since vue.js supports many other options, it may even be a good idea to put all these customised settings into a separate config file instead of injecting them all via env vars.

rhtyd:
Thanks @onitake, I'm on leave this week. I'll revisit this PR next week.

onitake:
@rhtyd Thanks to the help of a colleague who knows JS better, I found a way to make the additional options optional. Please give them a try!

rhtyd:
Manually tested, LGTM and made some changes. Manually squashed and merged the PR.

PR #8, opened 2019-10-28
Fix misused class in the info card component. Adjust margin.

iceteabottle:
I fixed the margin between the detail items of the info card component. I removed the imho misused css class.


iceteabottle:
Before:
before

After:
after

Issue #18, opened 2019-11-19
Large number of errors from eslint


onitake:
When building primate with eslint installed, a large number of errors are thrown.

Related: #17 

Full log: https://gist.github.com/onitake/d81c435697c023463865e8581d968e11

Excerpt:
```
Module Error (from ./node_modules/eslint-loader/index.js):

/build/src/components/header/HeaderNotice.vue
   64:9   error  Unnecessarily quoted property 'done' found      quote-props
   64:19  error  Unnecessarily quoted property 'icon' found      quote-props
   64:43  error  Unnecessarily quoted property 'style' found     quote-props
   65:9   error  Unnecessarily quoted property 'progress' found  quote-props
   65:23  error  Unnecessarily quoted property 'icon' found      quote-props
   65:42  error  Unnecessarily quoted property 'style' found     quote-props
   66:9   error  Unnecessarily quoted property 'failed' found    quote-props
   66:21  error  Unnecessarily quoted property 'icon' found      quote-props
   66:45  error  Unnecessarily quoted property 'style' found     quote-props
   88:46  error  Unnecessarily quoted property 'jobid' found     quote-props
   92:34  error  ["success"] is better written in dot notation   dot-notation
  103:34  error  ["error"] is better written in dot notation     dot-notation

✖ 12 problems (12 errors, 0 warnings)
  12 errors and 0 warnings potentially fixable with the `--fix` option.
```

These should be addressed.

rhtyd:
Send a PR @onitake or advise how you fixed those errors?

onitake:
I see two ways to "fix" these errors:

1. Apply the suggested fixes - which I'm not sure are safe. JS is a bit outside my domain.
2. Disable eslint.

onitake:
The errors are triggered by the `"@vue/standard"` extension in the `eslintConfig` section in package.json
If I remove that, Primate will build and run fine.

While I think that fixing the errors is better, it shouldn't be necessary to have the build fail because of them.

rhtyd:
I'm unable to reproduce them on latest master @onitake - does it fix to use the eslint4.x version?
```
> npm run lint

> primate@0.1.0 lint /home/rohit/lab/apache/primate
> vue-cli-service lint

 DONE  No lint errors found!
```

onitake:
I tried with eslint 4.0, but it still reported errors. Not sure if this was caused by leftovers or something else. I also did so in a fresh container, so I don't think it was on my host system...

[...]

Issue #17, opened 2019-11-19
Missing eslint peer dependencies


onitake:
I found some more dependencies that are missing:
* eslint-plugin-import
* eslint-plugin-node
* eslint-plugin-promise
* eslint-plugin-standard

Without these, the build step will fail.

rhtyd:
@onitake Did you run `npm install` on primate, after installing global packages - `sudo npm install -g @vue/cli webpack eslint`

onitake:
Yes, I did. But this only installs the tools globally and not the peer dependencies. Those need to be added to the package.json.

onitake:
When the deps are installed, eslint throws a ton of errors. I'll open a separate issue for that.

rhtyd:
Thanks for sharing, do you know (or have figured) how to fix your issues @onitake ?

onitake:
#19 now includes fixes for the lint errors.

PR #19, opened 2019-11-19
Add missing eslint dependencies and fix lint errors

onitake:
**Caveat:** Mutually exclusive with #20 

Fixes: #17 
Fixes: #18 

These dependencies are necessary or eslint will fail with the standard vue.js rule set.
The PR also includes fixes for all lint errors.


onitake:
The Travis build is now failing due to the eslint checks...

onitake:
Added a commit with all vue-standard mandated lint errors fixed.
Most of them were automatically applied by eslint, the [no-prototype-builtins](https://eslint.org/docs/rules/no-prototype-builtins) were fixed by hand.

:construction:  **Caveat:** The changes are still untested.

PR #20, opened 2019-11-19
Remove vue.js standard eslint rule set

onitake:
**Caveat:** Mutually exclusive with #19 .

May address #18 - but doesn't really fix it.

The vue.js standard rule set causes a lot of errors on the Primate source code.
This PR removes the standard rule set and only keeps the strongly-recommended rules.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,8 +87,7 @@
       "node": true
     },
     "extends": [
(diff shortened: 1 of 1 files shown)


rhtyd:
Will merge https://github.com/apache/cloudstack-primate/pull/19

Issue #12, opened 2019-11-11
Add instructions on how to avoid system-wide global npm package installation


onitake:
Commands like
```bash
sudo npm install -g @vue/cli webpack eslint
```
will install global dependencies system-wide, which will cause conflicts with files installed by a package manager (rpm, dpkg, ...).

There should be instructions in the README on how to avoid overriding system files.

For example, it's possible to set up a separate runtime directory for node/npm modules and simply run npm from there. This will pick up the local modules instead of those from the system.

rhtyd:
@onitake thanks, we're all new to UI so best practices are likely not known or followed but PRs are welcome.

onitake:
I'll try to write Dockerfile that illustrates how this should work.
So far, I've done my test deployments using Ansible, but having a Primate  container may be very useful for quick rollout.

PR #21, opened 2019-11-19
Add build dependencies to package.json and document production build properly

onitake:
Fixes: #12

With build tooling added to the dev dependencies, it is no longer necessary to install system-wide.
This is particularly useful for production builds, where system-wide package pollution is highly discouraged or even prohibited.

npm@next, vue-cli and ncu may still be useful to developers, but they shouldn't be part of the build instructions. A separate readme update is on the way.

The PR now also includes a readme update to clarify which tools should be installed when.
Production builds do not need global tool installs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,13 +15,11 @@ Install node: (CentOS/Fedora)
     curl -sL https://rpm.nodesource.com/setup_10.x | sudo bash -
     sudo yum install nodejs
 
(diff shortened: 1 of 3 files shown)


onitake:
Nice, I didn't even think about running Primate inside the Cloudstack mgmt server.
The jetty instructions are a valuable addition.

PR #23, opened 2019-11-20
Change CRLF editor config to LF

onitake:
Practically all files in the repository have UNIX (LF) line endings.
This contradicts the CRLF policy in .editorconfig and is causing problems with all editors that honour the editor config.

This PR replaces the CRLF policy with LF.


rhtyd:
Thanks @onitake - I personally only use vim, this was from the scaffolding/template boilerplate.

Issue #27, opened 2019-11-22
Implement Tree-view based Domain List View


rhtyd:
Implement a tree based domain list view, or discuss alternative approaches. (ref listDomainChildren)

The suggested component that can be used:
https://www.antdv.com/components/tree/
Or,
A table with a tree based navigation like: 
https://www.antdv.com/components/table/#components-table-demo-tree-data

Screenshot of how it looks in the old UI:
![Screenshot from 2019-11-25 23-26-36](https://user-images.githubusercontent.com/95203/69565495-156de700-0fdb-11ea-8bf3-caf3d21a1452.png)

This is how the domain list view looks in Primate:
![Screenshot from 2019-11-25 23-28-21](https://user-images.githubusercontent.com/95203/69565587-4e0dc080-0fdb-11ea-954f-237e99d0a4ae.png)

svenvogel:
Hi @rhtyd, my college Hoang Nguyen @utchoang will look in this task. Give him some time to come into the  topic.

rhtyd:
Sure, thanks for sharing @svenvogel let me know if you or @utchoang needs any assistance. I'll update tickets with details in coming days.

Issue #46, opened 2019-11-26
Account Certs Tab


rhtyd:
Add a new Certificates tab in the account detail view, it should support apis to list/add/delete (listSslCerts, uploadSslCert, deleteSslCert)

Reference screenshot from old UI:

![Screenshot from 2019-12-03 22-34-45](https://user-images.githubusercontent.com/95203/70072589-4a9eb880-161d-11ea-98f7-c49caaa4e68f.png)
![Screenshot from 2019-12-03 22-34-52](https://user-images.githubusercontent.com/95203/70072592-4a9eb880-161d-11ea-9c71-9ebc4d87765b.png)

Also check and compare and implement any missing detail or action by comparing the old UI vs the new UI, except for the settings tab which is done in #26

rhtyd:
I'll update the requirements/screenshots and details on this issue this week soon.

svenvogel:
old view

![image](https://user-images.githubusercontent.com/5433844/70060676-4b1a5d80-15e3-11ea-9e9a-4854420ae345.png)

rhtyd:
@svenvogel no, not that. The SSH keypair is already implemented and is under the compute navigation group now. I've updated the ticket with details now.

rhtyd:
Once this issue's PR is up, I'll add details on https://github.com/apache/cloudstack-primate/issues/41

utchoang:
@rhtyd @svenvogel 
This is the UI I sketched for accounts. 
Could you please review and give me your comments, especially for the group of domains
![account_tree](https://user-images.githubusercontent.com/13766648/70402010-7d551080-1a64-11ea-94b4-6f20b7f585d5.png)

svenvogel:
@rhtyd do you mean your pictures? we need to add the new certs tab right?

rhtyd:
@utchoang - no need for this as we've got your tree-based domain PR merged. What we want with this ticket is to allow users to upload certificates to their accounts. See the original issue description: https://github.com/apache/cloudstack-primate/issues/46#issue-528894680

utchoang:
@rhtyd I got it, then I will move to the cert tab

utchoang:
@rhtyd @svenvogel 
- Please review the UI for certificate tab and let me know your comments.
![certificate](https://user-images.githubusercontent.com/13766648/70502884-9b01a300-1b54-11ea-929b-2a8208a55e18.png)

[...]

Issue #26, opened 2019-11-22
Re-usable and editable Configurations


rhtyd:
- Global settings (editable column/keys), support list/update configuration APIs, pagination and searching
- Support reusable settings tab across accounts, domains, zones, clusters, primary&secondary storage
- APIs to be supported: listConfigurations and updateConfiguration
- Requirement: The component should be highly re-usable and implemented that can take in resource and loading as props and should detect which view it is in using `$route`. See #38  for example and reference only (don't extend that, implement a new component). Make suitable/minimal changes to `components/view/ListView.vue` and/or `AutogenView` to allow reusability as the default list/table view component as well.

![Screenshot from 2019-12-03 23-14-43](https://user-images.githubusercontent.com/95203/70075526-ff87a400-1622-11ea-9dac-cc62e01d894c.png)

Based on the `$route.name` it can (implement like a switch) it should pass relevant scoped ID such as a zoneid, accountid etc for both the list and update APIs. For example:

![Screenshot from 2019-12-03 23-16-05](https://user-images.githubusercontent.com/95203/70075504-f4cd0f00-1622-11ea-99f0-08162d275a58.png)

Examples:

Here is how you can create a custom Tab component:
https://github.com/apache/cloudstack-primate/blob/master/src/views/iam/RolePermissionTab.vue

Here's how you can define the tab components in the config file:
https://github.com/apache/cloudstack-primate/blob/master/src/config/section/iam.js#L223

Here's an example of creating a list, for/if etc directives usage:
https://github.com/apache/cloudstack-primate/blob/master/src/components/view/DetailSettings.vue

rhtyd:
Related for reference: https://github.com/apache/cloudstack-primate/pull/38

rhtyd:
Fixed in https://github.com/apache/cloudstack-primate/pull/63

PR #53, opened 2019-12-02
Issue #27: Implement Tree-view based Domain List View

utchoang:
Hi @rhtyd, @svenvogel. This is the design of the new tree-view. Please let me know your comments.
 
Domain_treeview
subitem_tree_selected


src/components/view/DetailsTab.vue
```diff
@@ -23,7 +23,8 @@
       <div>
         <strong>{{ $t(item) }}</strong>
         <br/>
-        <div>
+        <div v-if="resource[item]==='Unlimited'">-1</div>
```
rhtyd: @svenvogel @PaulAngus @borisstoyanov  do we want to show -1 instead of unlimited?

src/config/section/iam.js
```diff
@@ -176,6 +176,17 @@ export default {
...
+        {
+          name: 'details',
+          component: () => import('@/components/view/DetailsTab.vue')
+        }, {
+          name: 'settings',
+          permission: ['listConfigurations'],
+          component: () => import('@/views/iam/TemplateDomainSettings.vue')
+        }
+      ],
+      treeView: true,
```
rhtyd: This will require refactoring, however, I'll accept it.

rhtyd:
@utchoang @svenvogel here are some issues from my testing:
- On refreshing the view loses context and shows me the root domain, even if the nested domain may be selected (the refresh can happen if I add a new domain or perform an action like update count)
- The + plus button the tree should only be shown the domain has `haschild` set to true (no need to show it otherwise)
- In the src/config/section/iam.js for the createDomain action, you can add a mapping rule for `parentdomainid` (see compute.js for example) to return the parentdomain.id as the `record.id`, i.e. id of the selected domain
- As in the info-card component, can we show additional details about the domain, or at least the button to view account under that domain (see old UI)
- Would be great if the file icon could be replace with the `domain` icon in the tree if that's possible for leaf domains with no children
- Remove the settings tab for the account, as it will be dealt by #26 

Screenshots may help:

- Old UI (refreshing does not reset the view)
Screenshot from 2019-12-03 23-09-03

- Info card for a domain per `master` branch, see comments and the view accounts button:
Screenshot from 2019-12-03 23-08-46

- screenshot that shows different in the list of rendered keys:
Screenshot from 2019-12-03 23-04-14
Screenshot from 2019-12-03 23-05-01

src/config/section/iam.js
```diff
@@ -176,6 +176,17 @@ export default {
         title: 'Accounts',
         param: 'domainid'
       }],
+      tabs: [
```
rhtyd: @utchoang @svenvogel: instead of defining a new treeview, you can actually define the component that should render the resource/list view by saying something like:
```
  ...
  component: () => import('@src/components/view/TreeView.vue'),
  ...
```
I'm also fine with current implementation if doing the above change requires a lot of effort.

src/views/AutogenView.vue
```diff
@@ -309,6 +326,12 @@ export default {
         this.dataView = false
       }
 
+      if (this.$route && this.$route.meta && this.$route.meta.treeView) {
```
rhtyd: @utchoang @svenvogel rewrite this as:
```
  this.treeView = this.$route && this.$route.meta && this.$route.meta.treeView
```

src/views/iam/TemplateDomainSettings.vue
```diff
@@ -0,0 +1,118 @@
...
+    </a-spin>
+  </div>
+</template>
+
+<script>
+import DetailSettings from '@/components/view/DetailSettings'
+import { api } from '@/api'
+
+export default {
+  name: 'TemplateDomainSettings',
```
rhtyd: @utchoang @svenvogel this can be removed as we'll have the settings (scoped) implemented via #26

utchoang:
@rhtyd Thanks for your response. I will check and fix it.

src/components/view/DetailsTab.vue
utchoang: correct, but I revised the display data and removed it as it is now not necessary

src/config/section/iam.js
utchoang: Yes, it will take lots of effort. I think we can go that way later, when we have time and have already completed other tasks with higher priority

src/views/AutogenView.vue
utchoang: Yes, I rewrite it

src/views/iam/TemplateDomainSettings.vue
utchoang: I removed this component.

src/config/section/iam.js
utchoang: I think it is similar to https://github.com/apache/cloudstack-primate/pull/53#discussion_r353568566

utchoang:
@rhtyd @svenvogel It is fixed. Please check. Regarding to the info card, I'm going to make a tab as in the image below. Please let me know what you think
Screenshot_1

rhtyd:
Thanks @utchoang I'll review and merge, maybe post merging do few fixes.

rhtyd:
@utchoang @svenvogel Overall good work!

Here a few outstanding issues:

- Go to a resource that links to a domain, for example a VM -> click on the domain -> it takes you to the domain resource view (path is something like http://localhost:5050/#/domain/d150293a-eb52-11e9-b6c2-1e00970107cd). Here we see two infocards.
- There could be a middleground to 
- On clicking a domain the URL router does not change, maybe if this is implemented as a view component (instead of a new tree component) it would simplify a lot -> let me know if you think it would be a lot of work for you I can takeover to complete it

Screenshot from 2019-12-05 23-01-54

utchoang:
@rhtyd Thank for your response. I made change so that the view of the domain resource will not display domain's two info cards

svenvogel:
>Here a few outstanding issues:

Go to a resource that links to a domain, for example a VM -> click on the domain -> it takes you to the domain resource view (path is something like http://localhost:5050/#/domain/d150293a-eb52-11e9-b6c2-1e00970107cd). Here we see two infocards.

There could be a middleground to

On clicking a domain the URL router does not change, maybe if this is implemented as a view component (instead of a new tree component) it would simplify a lot -> let me know if you think it would be a lot of work for you I can takeover to complete it

@rhtyd lets us change alls things completely so its ready for merge... 

@utchoang there are things to do?

rhtyd:
alright let me handle and takeover this, I think if we reimplement this as a resource view component we can handle all cases. @utchoang in that case you can move to the next task/ticket like #46 etc. and grant me push access to your PR repo. I'll refactor and merge this, this weekend.

rhtyd:
I've made some fixes @utchoang @svenvogel and will merge this. Any further changes or fixes can be done in a new PR.

PR #63, opened 2019-12-09
Reusable settings

RitchieVincent:


src/components/view/ListView.vue
```diff
@@ -22,7 +22,6 @@
     :columns="columns"
     :dataSource="items"
     :rowKey="record => record.id || record.name"
-    :scroll="{ x: '100%' }"
```
RitchieVincent: This property was removed to allow for horizontal scrolling on mobile devices when the table becomes too wide for the viewport.

PR #67, opened 2019-12-11
VM Migration Action form

RitchieVincent:


rhtyd:
LGTM, did not test actual API action. Can you adjust the CSS and size of the popup, if possible?
@RitchieVincent 

Screenshot from 2019-12-11 23-03-04

RitchieVincent:
> LGTM, did not test actual API action. Can you adjust the CSS and size of the popup, if possible?
> @RitchieVincent
> 
> Screenshot from 2019-12-11 23-03-04

Is that how you want it to look, in one row rather than in a 2x2 grid like it is at the moment?

rhtyd:
@RitchieVincent I would prefer one host on each row and the checkbox/select button either on left or right (see the screenshot in old UI, we've each host on one single row). Also if the form size (width and lenght) could be slightly bigger. Rest looks okay.

RitchieVincent:
> @RitchieVincent I would prefer one host on each row and the checkbox/select button either on left or right (see the screenshot in old UI, we've each host on one single row). Also if the form size (width and lenght) could be slightly bigger. Rest looks okay.

No problem, I'll update it now.

RitchieVincent:
@rhtyd The requested changes have been completed now.

rhtyd:
@RitchieVincent please see the minor fixes I've done. Notice and try to use the translation using `{{ $t('some string') }}` as much as possible in your work, so as to avoid a refactoring in future. You can check is existing string exist under the `src/locale/en.json`, or you may even port the old strings (at least english) from https://github.com/apache/cloudstack/blob/master/ui/l10n/en.js. The only thing to keep in mind is to keep the `key` of the translation string simple (for example instead of label.vm, just vm).

I'll leave few comments on the code for you to refactor.

src/views/compute/MigrateWizard.vue
```diff
@@ -36,12 +85,146 @@ export default {
...
+        virtualmachineid: this.resource.id
+      }).then(response => {
+        this.pollActionCompletion(response.migratevirtualmachineresponse.jobid)
+      }).catch(error => {
+        console.error(error)
+        this.$message.error('Failed to migrate host.')
+      })
+    },
+    pollActionCompletion (jobId) {
+      api('queryAsyncJobResult', { jobId }).then(json => {
```
rhtyd: @RitchieVincent This seems to be fairly commonly used method/pattern, can you consider refactoring this out to a utility method that we can use as a mixin (see AutogenView or other components for example on using mixins, the utils/mixin.js is one example).

src/views/compute/MigrateWizard.vue
```diff
@@ -36,12 +85,146 @@ export default {
...
+        if (result.jobstatus === 1) {
+          this.$message.success(`Migration completed successfully for ${this.resource.name}`)
+          this.$parent.$parent.close()
+        } else if (result.jobstatus === 2) {
+          this.$message.error(`Migration failed for ${this.resource.name}`)
+          this.fetchData()
+        } else if (result.jobstatus === 0) {
+          this.$message
+            .loading(`Migration in progress for ${this.resource.name}`, 3)
+            .then(() => this.pollActionCompletion(jobId))
```
rhtyd: How about we close the form, irrespective of success or failure like others for consistency? Alternatively, at least on errors show the full error message from the job response?

rhtyd:
@RitchieVincent one more thing I forgot, we probably also want some form of pagination. You may check against the old UI. Thanks.

src/views/compute/MigrateWizard.vue
RitchieVincent: I've now created a reusable method which can be used for this in `utils/methods.js`.

RitchieVincent:
> @RitchieVincent one more thing I forgot, we probably also want some form of pagination. You may check against the old UI. Thanks.

I thought having pagination inside the modal would be a bad idea in terms of UX. Instead I've set it up to scroll if there are too many items. If we want pagination I'd need a couple of hours, as I'd need to rebuild parts of this component to allow for it. Let me know if you want me to go ahead with it.

rhtyd:
@RitchieVincent maybe at least call the api with a large pagesize like 500, we can revisit implementing pagination in future?

rhtyd:
I went ahead with some changes, please follow the changes @RitchieVincent

PR #71, opened 2019-12-16
VM Hardware: NICs and IPs management 

RitchieVincent:
#58

PR #72, opened 2019-12-16
Volume storage migration

RitchieVincent:


RitchieVincent:
@rhtyd Just to note, I've added the exact same file/method located in `src/utils/plugins.js` as PR #71 .

src/utils/plugins.js
```diff
@@ -0,0 +1,76 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+import { api } from '@/api'
+import { message, notification } from 'ant-design-vue'
+
+export const pollJobPlugin = {
+
+  install (Vue) {
+    Vue.prototype.$queryAsyncJobResult = function (options) {
```
rhtyd: I think the rename was reverted @RitchieVincent - change it back to something small like `$pollResult` or `$pollJob`

RitchieVincent: Will do now

PR #73, opened 2019-12-18
[WIP] Project: Project View and Actions

utchoang:
- [x]  Project view and Project dashboard
- [x]  Projects: delete/list project accounts etc.
- [x]  Project invitations: list/delete/updat
- [x] Project resource tab in the project detail view


utchoang:
@rhtyd @svenvogel 
This is the design of the Project View and Actions. Please let me know your comments.
1. Project dashboard
project_dashboard_1
project_dashboard_2
project_dashboard_3

2. Projects account:
project_account_1

3. Project invitaion:
project_invitation_1
project_invitation_2
project_invitation_3

rhtyd:
@utchoang couple of issues: cc @svenvogel 
- account tab not show after I create a project, and see the project's detailview
- how do I add/update project invitation by token (updateProjectInvitation? the old UI allow entering project ID and token)
- in the project dashboard's account tab, the account list does not update after adding a new account
- Can you use `this.$parent` etc. to find and add the badge to the project invitation, see if it's feasible to avoid changing the AutogenView too much without requiring customisation in the general purpose component?

Can you address those, I'll do a thorough review later tomorrow.

svenvogel:
@utchoang cc @rhtyd Rohit wrote the technical things from me only a design suggestion...

we should a add more filter and an search bar
image

i am also thought about tags but they are not so useful i think

rhtyd:
@utchoang Along with the raised issues that you'll address, kindly also add the missing resources tab in the project detail view that allows the admin/owner to configure the limits of resources allowed for that project: 

Screenshot from 2019-12-19 04-25-22

utchoang:
> @rhtyd cc @svenvogel
> 
> * how do I add/update project invitation by token (updateProjectInvitation? the old UI allow entering project ID and token)

In the old UI that I got, I cannot find the option to add token when accepting or declining an invitation. Could you please send me the UI where the token could be added?

rhtyd:
@utchoang this is on 4.13 version:

In the project invitation list view the Enter token button is seen here:
Screenshot from 2019-12-19 15-50-21

Clicking it opens this dialog:
Screenshot from 2019-12-19 15-50-12

Upon entering the details, the following API call is made:
```
command: updateProjectInvitation
response: json
projectid: some-uuid
token: some-token
_: 1576750780606
```

rhtyd:
I did another test, the `resources` tab to set the limits is only shown to admin @utchoang

rhtyd:
@utchoang fyi - I've merged latest master (was doing some tests), will continue the review/test once you can confirm the fixes. Thanks.

rhtyd:
@utchoang update - I've fixed permissions issues now, that should fix the issue of when to show the tabs or the action button, and added feature discovery which can be used to determine the: cc @svenvogel 
```
    "allowusercreateprojects": true,
...
    "projectinviterequired": true,
```
Kindly rebase the branch.

utchoang:
@rhtyd cc @svenvogel : I've completed the fixes:
- In the popup to add/update project invitation by token, the Autogenview is set to get the field type of projectid as listbox, thus, I've changed the project type to input, not listbox.
mapping


@svenvogel https://github.com/apache/cloudstack-primate/pull/73#issuecomment-567068116
I think there are some problems for this issue as below
- The parameter keyword of API listProjectInvitations seems not to work. I did try to add/change keyword but the search results did not change accordingly
- Additional filter: do you mean you'd like to have multiple filters for column project and domain? If multiple filters are used, the pagination would not work correctly. The reason is the API does not support search project list and domain list. Please check and let me know your idea.

rhtyd:
Alright I'll review this in a while @utchoang - for the mapping, the allowed type are limited (I'll document that later on). To get automatic listing of projects, the projectid mapping shouldn't be defined.

rhtyd:
@utchoang quick review comments:
- The resource tab under projects view does not work where input buttons are not seen, however it works via dashboard view:
Screenshot from 2019-12-20 11-44-10
- The badge on action does not work or show pending invitations 
- User dashboard (when not in a project, but default view) is broken

utchoang:
@rhtyd Sorry for this problem. When I merged from the master, it was lost.

rhtyd:
@utchoang sure no problem, I'll review and get back to this PR shortly

rhtyd:
@utchoang can you perform an end-to-end testing of all the features/cases. I found a few you can try to fix:
1. Non-admin user's dashboard no longer shows the cards:

Screenshot from 2019-12-20 12-19-14

2. When owner of a project switches to a project, the account tab is not visible: (this is a non-admin user account)
Screenshot from 2019-12-20 12-19-40

3. The pending invitation badge is not shown? (perhaps a merge issue)

I've merged latest master, kindly rebase and continue testing/fixing. Thanks.

utchoang:
@rhtyd
* I have fixed for https://github.com/apache/cloudstack-primate/pull/73#issuecomment-567812849
* for https://github.com/apache/cloudstack-primate/pull/73#issuecomment-567788261, I understand and I make change by importing another component

src/components/view/ListView.vue
```diff
@@ -229,6 +232,54 @@ export default {
     editValue (record) {
       this.editableValueKey = record.key
       this.editableValue = record.value
+    },
+    handleShowBadge () {
```
rhtyd: This logic should ideally be moved to the action component (instead of bloating the list view component)

src/config/router.js
```diff
@@ -167,7 +167,29 @@ export const asyncRouterMap = [
...
+              name: 'accounts',
+              show: (record, route, user) => { return record.account === user.account || ['Admin', 'DomainAdmin'].includes(user.roletype) },
+              component: () => import('@/views/projects/AccountsTab')
+            },
+            {
+              name: 'resources',
+              permission: ['updateResourceLimit'],
+              args: ['network', 'volume', 'public_ip', 'template', 'user_vm', 'snapshot', 'vpc', 'cpu', 'memory', 'primary_storage', 'secondary_storage'],
+              show: (record, route, user) => { return ['Admin'].includes(user.roletype) },
+              component: () => import('@/views/projects/ResourcesTab.vue')
```
rhtyd: Since this is very specific to whether we're in a project view or not; this maybe moved to the dashboard component itself.

src/views/AutogenView.vue
```diff
@@ -45,17 +45,31 @@
               <template slot="title">
                 {{ $t(action.label) }}
               </template>
+              <a-badge
```
rhtyd: Let's see if we can find a way to avoid bloating this component

src/views/AutogenView.vue
```diff
@@ -291,7 +312,9 @@ export default {
       actions: [],
       treeData: [],
       treeSelected: {},
-      actionData: []
+      actionData: [],
+      actionBadge: [],
+      adminType: 'Admin'
```
rhtyd: This looks like unused variable, if so please remove.

src/views/dashboard/UsageDashboardChart.vue
```diff
@@ -0,0 +1,63 @@
...
+    },
+    loading: {
+      type: Boolean,
+      default: false
+    }
+  }
+}
+</script>
+
+<style scoped>
```
rhtyd: Remove style if no class is declared.

rhtyd:
Overall LGTM, except for few edge case issues (like refresh inside project resource view fails when account tab is opened). It needs a bit of refactoring, and allowing projectid to be any kind of text rather than a select dropdown only. Good effort @utchoang, if you've time please work on the raised issues, otherwise try and handle those next week.

svenvogel:
>Additional filter: do you mean you'd like to have multiple filters for column project and domain? If multiple filters are used, the pagination would not work correctly. The reason is the API does not support search project list and domain list. Please check and let me know your idea.

@rhtyd you are right. hmm okay let me think about that. can we use "tags" i mean add a tag "domain with...." and a "tag with project name ...."?
the search bar should not be a problem. right?

utchoang:
> Overall LGTM, except for few edge case issues (like refresh inside project resource view fails when account tab is opened). It needs a bit of refactoring, and allowing projectid to be any kind of text rather than a select dropdown only. Good effort @utchoang, if you've time please work on the raised issues, otherwise try and handle those next week.

@rhtyd Could you please explain the last idea about projectid a little more? The field for projectid could be either text input or dropdown selection. Do you mean to allow input search by projectid?

rhtyd:
Hi @utchoang ignore my last comment around the projectid/input type. But follow my last change (around handling changing resource to handle fetchData/refreshing the view). Can you perform an end-to-end functional testing of the feature once? I've made some changes and I think it's ready for merging after a final round of testing and review. (I'm doing code re-review, but LGTM from my side on functionality).

utchoang:
Hi @rhtyd The issues mentioned in PR are fixed and the refactoring is also done. Functional testing is now completed. Please check.

Hi @svenvogel 
* https://github.com/apache/cloudstack-primate/pull/73#issuecomment-568190912 Regarding to the tags as in your idea, I have the below suggestion
filter_domain

* The search bar: The parameter keyword of API listProjectInvitations seems not to work. I did try to add/change keyword but the search results did not change accordingly

svenvogel:
>Additional filter: do you mean you'd like to have multiple filters for column project and domain? If multiple filters are used, the pagination would not work correctly. The reason is the API does not support search project list and domain list. Please check and let me know your idea.

@utchoang - @rhtyd wrote this above. i dont know if this works?

good suggestion.there are any questions for me and maybe we can change it like this.

1. does this tags working?
2. if we use tags and they are working should we remove the filter from state and use only tags for "domain" and "filter". what i understood from @rhtyd was if the api dont give us a filter element it will not work. @rhtyd is this true?
3. i think tags and domain will be the same and maybe we should not do that. i thought like this.
https://ant.design/components/select/
image
can we use it as one element "filter" with tags and "search" bar? is this possible or do we need to get it from the api?

utchoang:
> > Additional filter: do you mean you'd like to have multiple filters for column project and domain? If multiple filters are used, the pagination would not work correctly. The reason is the API does not support search project list and domain list. Please check and let me know your idea.
> 
> @utchoang - @rhtyd wrote this above. i dont know if this works?
> 
> good suggestion.there are any questions for me and maybe we can change it like this.
> 
> 1. does this tags working?
> 2. if we use tags and they are working should we remove the filter from state and use only tags for "domain" and "filter". what i understood from @rhtyd was if the api dont give us a filter element it will not work. @rhtyd is this true?
> 3. i think tags and domain will be the same and maybe we should not do that. i thought like this.
>    https://ant.design/components/select/
>    image
>    can we use it as one element "filter" with tags and "search" bar? domain and state together? @rhtyd  is this possible or do we need to get it from the api?
@svenvogel cc @rhtyd 
1. As what I found from the source code http://cloudstack.apache.org/api/apidocs-4.13/apis/listProjectInvitations.html, that the listProjectInvitations requests "keyword" parameter seems not to work. I tried with different value and it always returns the same result.

2. If requesting "keyword" parameter works, filter based on selecting tags won't be needed as we can search filters by entering domain name or project name.

rhtyd:
@utchoang it's possible that some list APIs don't honour or implement the keywords based searching, in which case you may file a bug here: https://github.com/apache/cloudstack/issues/new
Agree on the tags comment cc @svenvogel - maybe we can revisit this once the searching is supported. I think I can live with the current implementation and iterate in future.

I've started reviewing/testing it now, hopefully will merge in next few hours.

svenvogel:
@rhtyd @utchoang okay. lets leave it without tags. if its implemented in future we can add this. lets move to the next one.

rhtyd:
@utchoang LGTM, can you address the following issues in this PR and other PRs:
- Whenever any action/button needs to be added in a custom action/form/tab, always check if the user has access to execute the API using `apiName in this.$store.getters.apis` or similar
- Deleting a user/account from a project, or making another account the project admin does not cause refreshing of the accounts tab

Without blocking this further I'll merge this, but please send another PR to fix issues in this PR @utchoang  Thanks.

PR #89, opened 2019-12-27
[WIP] - Register ISO Action

utchoang:
* [x]  ISO register with multi-zone by URL
@rhtyd @svenvogel 
This is the design of the Custom Register ISO action. Please let me know your comments.
Untitled


rhtyd:
Looks good, let me test and get back to you by Monday @utchoang - meanwhile you may explore reusing the code/component as most of the fields would be same as that generated by the auto-generated form except for the zoneid select field.

src/views/image/RegisterIso.vue
```diff
@@ -0,0 +1,231 @@
...
+            default:
+              params[key] = input
+              break
+          }
+        }
+
+        this.loading = true
+
+        api('registerIso', params).then(json => {
+          console.log(json)
```
rhtyd: @utchoang here we can call `this.$pollJob` to check/poll for the registerIso task.

rhtyd: Ignore just checked, the API is not async API.

PR #92, opened 2020-01-02
Register Template Action

utchoang:
@rhtyd cc @svenvogel 
This is the design of the Custom Register Template action. Please let me know your comments.
Screenshot_5
Screenshot_6
Screenshot_7
Screenshot_8
Screenshot_9
Screenshot_10
Screenshot_11


src/config/section/image.js
```diff
@@ -50,7 +50,11 @@ export default {
           icon: 'plus',
           label: 'Create template',
           listView: true,
-          args: ['url', 'name', 'displaytext', 'directdownload', 'zoneids', 'hypervisor', 'format', 'ostypeid', 'checksum', 'isextractable', 'passwordenabled', 'sshkeyenabled', 'isdynamicallyscalable', 'ispublic', 'isfeatured', 'isrouting', 'requireshvm']
+          popup: true,
+          show: (record, user, apis) => {
+            return Object.prototype.hasOwnProperty.call(apis, 'registerTemplate')
```
rhtyd: This is not needed for actions that are rendered in the top bar near breadcrumbs because the actionbutton.vue already has: `v-if="action.api in $store.getters.apis ...`

rhtyd:
@utchoang cc @svenvogel  - thanks for this, it's a slightly complex form, I think it needs a bit of testing. In my quick testing some of the models are tied, for example when selecting xenserver I'm unable to uncheck the XS-tools 6.1+ version option. Could you run an end to end testing?

utchoang:
@rhtyd I will test again and there will be a commit later. Thank you

rhtyd:
@utchoang one more point - the form is rather very long, could the form elements be clubbed in the same row, especially when the elements may be smaller (taking less width)? I don't have any specifics, maybe @svenvogel can comment?

utchoang:
@rhtyd cc @svenvogel In this regard, I think I will not use pop up, but will go to a new route, what do you think?

rhtyd:
No I think popup is fine @utchoang, since it's a single API (does not involve a very complex form). All I'm suggesting is clubbing multiple inputs (inputs, radio/checkbox) within same row if it looks alright (right now, all the fields are in a new line or row).

utchoang:
@rhtyd cc @svenvogel The switch section I have 2 suggestions:
Option 1:
Screenshot_16

Option 2:
Screenshot_17

What do you think about my suggestion.

rhtyd:
@utchoang while option1 looks good, I think let's go with option2 to be consistent with the other forms. I also dont' mind if you do two checks/sliders per row instead of three. @svenvogel what do you think?

@utchoang also check what options to show if the user is not a root admin? For example, this is what I see as a normal user:
Screenshot from 2020-01-03 11-50-58
You can check the user's role type using: `this.$store.getters.userInfo.roletype`.

utchoang:
@rhtyd cc @svenvogel This is my new design. What do you think?
1. Default


2. Xenserver (hypervisor)


3. KVM (hypervisor)


4. VMW (hypervisor)

svenvogel:
@utchoang i would prefer more the first one. they looks smaller and not so heavy. @rhtyd what do you think?

svenvogel:
@utchoang there is a little margin they should not there. can we fix this so that we have a nice edge? (left and right side)

image

utchoang:
@svenvogel, This problem I've fixed in https://github.com/apache/cloudstack-primate/pull/92/commits/ed2285086f83ce90887ed29302d733f62f82017e
The current UI
Untitled

rhtyd:
Thanks @utchoang - is this ready for review and testing?

utchoang:
@rhtyd Yes. Please review it.

rhtyd:
@utchoang I've completed the review, made some fixes. However, the API request is not correct, for example all the extra options should be passed as `details`.
For example, to register a VMware template with custom nic, storage controller, nic controller, this PR send this request:
Screenshot from 2020-01-14 11-43-56

And, the old UI sends this request:
Screenshot from 2020-01-14 11-43-30

@utchoang please refer to the API doc here: http://cloudstack.apache.org/api/apidocs-4.13/apis/registerTemplate.html
... and for reference please refer to the old code/implementation here:
https://github.com/apache/cloudstack/blob/master/ui/scripts/templates.js#L123
cc @svenvogel

utchoang:
@rhtyd Thanks for your respond. I have reviewed and fixed the params of the API. Please review again
cc @svenvogel

rhtyd:
@utchoang can you do another round of testing, it's not the KVM or VMware, all additional information of a template such as the disk controller, nic controller, keyboard and XenServer tool version (for example in case of xenserver) are all sent as `details` of the API request. Please ping me once you've checked all those params against the old implementation.

utchoang:
@rhtyd I've already tested all. I'm sorry if the message in commit is unclear to you. Please review.

rhtyd:
Will review shortly @utchoang

PR #95, opened 2020-01-06
config: offerings list recursively

shwstppr:
Existing code was not listing *offerings recursively due to which domain admin user was not being able to see all offerings in its domain. This PR fixes listing behaviour.
Existing behaviour,
Screenshot from 2020-01-06 13-02-59

On old UI,
Screenshot from 2020-01-06 11-03-21

After changes,
Screenshot from 2020-01-06 13-02-01


shwstppr:
@rhtyd please review

PR #97, opened 2020-01-06
Security groups - Ingress + Egress tabs

RitchieVincent:


rhtyd:
@RitchieVincent merged, kindly follow this change. In most views, the resource already has the item, and we can reuse it in tabs and custom actions without calling the list API again. Kindly follow the change for future PRs:
https://github.com/apache/cloudstack-primate/pull/97/commits/736b720a4d76af84723325330e32de50cb370234

PR #99, opened 2020-01-09
Recurring Volume Snapshot Custom Action Form

utchoang:
@rhtyd cc @svenvogel
This is the design of the Recurring Volume Snapshot Action. Please let me know your comments.

* Schedule:

* Scheduled snapshots:


rhtyd:
Great, thanks @utchoang I'll review and get back to you by end of tomorrow.

svenvogel:
@utchoang from design looks good to me. if @rhtyd give okay after test we can merge it ... good work!

rhtyd:
@utchoang reviewed, LGTM except for this few issues:
1. Please always checkout new work/branch from origin/master, otherwise the PRs will have commits from other PRs/branches and will cause difficulties in reviewing them. I've fixed it manually for at least two PRs for now.

2. When I open the recurring action form the first time, I'm unable to list the timezones:
Screenshot from 2020-01-14 13-03-55

I've moved the timezone file/folder under utils and changed the import paths. It's also okay if instead of resolve/importing the json, you want to consume it as a static object/map.

3. While you're at it, can you also fix the createSnapshot form which takes in tags, which is not supported out of the box:
Screenshot from 2020-01-14 13-10-53

utchoang:
@rhtyd 
1. I have merged with the last master.

2. Timezone is not shown in the first place because I only allow it to be shown when the search text is entered. Reference: https://www.antdv.com/components/select. A lot of timezones listed could cause lag when entering other text inputs

3. I will fix and update it on the next commit

rhtyd:
@utchoang but it breaks the UX for the user, as in the old UI we would list all the possible select options when the form is loaded.

utchoang:
@rhtyd I would allow them to be all shown when the form is loaded, but it will affect other input (lag)

rhtyd:
@utchoang there are some conflicts, can you merge latest master and fix the conflicts?

utchoang:
@rhtyd I resolved the conflict PR. Please review again. Thank you

PR #103, opened 2020-01-13
config: fix existing access values display for update offering access

shwstppr:
This PR fixes showing existing domain and zone values in update offering access form.
Screenshot from 2020-01-06 13-19-57


shwstppr:
@rhtyd fixed the showing of original values on form open. Please review

PR #113, opened 2020-01-15
Dedicated domain reusable component

RitchieVincent:


rhtyd:
@RitchieVincent ping me when this has been refactored out as a separate re-usable dedicated component (at least for the action form)?

rhtyd:
@RitchieVincent there were some reactivity issues, please follow my changes and how `watch` could be used to solve such issues.

PR #119, opened 2020-01-20
Static nat form

RitchieVincent:


rhtyd:
@RitchieVincent can you merge latest master and check that when static nat is enabled for an IP; then only firewall tab must be shown.

RitchieVincent:
> @RitchieVincent can you merge latest master and check that when static nat is enabled for an IP; then only firewall tab must be shown.

Will do now 👍

RitchieVincent:
@rhtyd Only the firewall tab will show if Static NAT is enabled now.

rhtyd:
Thanks @RitchieVincent will review now

PR #127, opened 2020-01-22
Vmware dc section in details tab

RitchieVincent:
@rhtyd I made some minor changes to the `src/components/view/DedicateData.vue` to display in a nicer way in a list with this component. As we've discussed, I'm unable to run any of the API commands, but I did use some dummy data to test that it displays correctly, and that the create/update/delete buttons all show/hide correctly.
I've left it to use the autogen component to create the forms like you mentioned in the ticket, as it all works really well and has saved a lot of time recreating it all.


src/components/view/DetailsTab.vue
```diff
@@ -29,17 +29,19 @@
       </div>
     </a-list-item>
     <DedicateData :resource="resource" v-if="dedicatedSectionActive" />
+    <VMWare :resource="resource" v-if="this.$route.meta.name === 'zone'" />
```
rhtyd: @RitchieVincent in template, variables can be accessed without using `this.`.

rhtyd: Also, didn't we discuss to include VMware section as a separate Tab in the zone resource view. That means not embedded within the detailstab component, but as a separate tab component.

PR #138, opened 2020-01-30
VPC: Network ACL List Management

RitchieVincent:


rhtyd:
@RitchieVincent couple of issues with the PR:
- Make the list of items in a single row (use a-table if possible without borders, or use appropriate flex/grid?)
- Edit form does not fill all the values of the row
- Exported CSV does not match the data, as the csv from the old UI

PR #7, opened 2019-10-28
[WIP] Feature/vm wizard

iceteabottle:
- [x] Add selection for disk offering
- [ ] Format Info card for vm config summary
- [x] Add iso selection
- [x] Add tags to the templates (featured, public, etc.)
- [x] Sort templates/isos
- [x] Pin featured templates/isos at the top of each list
- [x] Add search filter input for isos/templates
- [x] Handle form change on switching from template to iso (and vice versa)
- [ ] Add translations for all input placeholders
- [ ] Implement submit method
- [x]  Add affinity group selection
- [ ] Add network selection
- [x] Add ssh key pairs selection
- [ ] Handle Navigation (e.g. reset the form on navigating away)
- [ ] Add filter possibility to selections (compute offering, disk offering, network offering, ...)

PR #146, opened 2020-02-05
Feature/vm wizard

utchoang:
* [x]  Add selection for disk offering
* [x]  Format Info card for vm config summary
* [x]  Add iso selection
* [x]  Add tags to the templates (featured, public, etc.)
* [x]  Sort templates/isos
* [x]  Pin featured templates/isos at the top of each list
* [x]  Add search filter input for isos/templates
* [x]  Handle form change on switching from template to iso (and vice versa)
* [x]  Add translations for all input placeholders
* [x]  Implement submit method
* [x]  Add affinity group selection
* [x]  Add network configuration (ip, mac) to network list params
* [x]  Add ssh key pairs selection
* [x]  Handle Navigation (e.g. reset the form on navigating away)
* [x]  Add filter possibility to selections (compute offering, disk offering, network offering, ...)
* [x] Change layout from collapse to steps
* [x] Add pod, cluster and host in step Basic Setup


rhtyd:
Thanks for submitting @utchoang 

@svenvogel cc @PaulAngus @borisstoyanov - from internal discussions one feedback I (got from Alex) is that the VM deployment wizard should not allow creation of new networks which may be confusing for users (multiple ways to do the same thing), instead we can add a help button or text that allows users to go to the network page and create the network they want themselves. This would however remove the previous UX that allowed users to add network while they deploy a VM. Thoughts?

utchoang:
@rhtyd thank you. I will add a Help button for creating new network while avoiding content switch.

andrijapanicsb:
+1 @rhtyd for network stuff during vm deployment wizard, but as suggested, please add some button so that user can still create a network.

utchoang:
@rhtyd I think I will add a Help button for creating new network while avoiding content switch

svenvogel:
@rhtyd i my experience most of my infra user find a way for creating a network in the wizard in the process of creating a virtual machine a good way. do complete creation in one step. my most users here was confused in the old UI because they dont know "whats that for network type and where is it connected"... if this is better in new UI i think it will not confuse.
 
i see 2 ways.

1. like @utchoang suggest i would use a help button if we dont want creation of network or "we have an ugly content switch". this means we avoid switch the site to another and back. this is from design perpective and usability not nice for handling. at the moment we have a speedy nice one pager vm wizard. each content switch will destroy it. help button like create network there... but not switch them...

2. for me +1 for the possibility to create directly the network because we use collapse mode. users they dont want it will not use it and if you plan network before you will go directly to network tab.

@rhtyd what do you think?

rhtyd:
@svenvogel I don't like the collapse mode, I would prefer all sections are visible. I'm okay either way to provide an option for users to create network. Maybe the network list/table can also say things like the type of the network, cidr etc. Help button is also fine. We may even reuse the create network component here to allow users to create a network, if they don't want to use an existing one.

There are some organisations that prefer network to be created, others don't. In most cases, say when deploying a VM (whether on cloud, or on desktop/server like VirtualBox or virt-manager) users usually are not allowed to create a network. For many providers the existing list of networks is like a network offering available to VMs, and users are generally only allowed to pick one of the existing ones (than create a new one). We can use router-link to allow users create networks for example?

utchoang:
@rhtyd cc @svenvogel There is error 503 found in http://acs.shapeblue.com:8080/client/. Please check

utchoang:
@rhtyd cc @svenvogel
This is my new design of the layout collapse mode to steps mode. Please review and let me know your comments.

utchoang:
@rhtyd cc @svenvogel 
Add pod, cluster and host in step Basic Setup

rhtyd:
@utchoang that's fine for the first iteration, I think as we had initially discussed I would rather prefer that all the views/sections are visible by default instead of steps/pages like the old UI's wizard. Perhaps once a user inputs say all the params of one step, the subsequent step may automatically show up?

In summary, let's just see if we can finish the first iteration and then seek feedback/reviews.

svenvogel:
@utchoang i agree with @rhtyd ... please combine the nice 1..2..3 blue step line but WITH one pager (scrollable in each section) ... "no next button". it is easier for user to change. maybe we had an misunderstood.

utchoang:
@rhtyd cc @svenvogel I have changed the following. Please let me know your comments.

utchoang:
Regarding to "Format Info card for vm config summary" with the image below, I still have no idea of what should be done. Could you please explain?

rhtyd:
Nice @utchoang - is this complete, ready for review/testing?

rhtyd:
@utchoang cc @svenvogel some review comments from testing the current branch:
- Maybe template and ISO could be show in tabs instead of collapse?
- Minor: show search bar like other steps for template/iso; the search bar could be above near to the place that may render on the right side of tabbed template/iso options
- Once I accidently select disk offering, no way to de-select it. Give a radio button to say no disk offering etc.
- Initially on the network step, if I select two networks, both are selected as default by the radio button
- Given option (radio button) to not select any ssh key (if we select one, there is no way to de-select) 
- Minor nit: the launch button icon uses poweroff icon, instead maybe use a play or rocket icon?

utchoang:
@rhtyd cc @svenvogel 
I have finished fixing the issues. Please review and give me your comment.
image
image

rhtyd:
@utchoang cc @svenvogel - is it ready for testing/review?

utchoang:
@rhtyd cc @svenvogel Yes. Please review it.

rhtyd:
@utchoang can you merge latest master on it?

utchoang:
@rhtyd Yes, I did merge the last master

rhtyd:
Great @utchoang, I'll start review/testing.

svenvogel:
@rhtyd CC @utchoang should we add groups, keyboard layout and userdata?

image

rhtyd:
@utchoang please pull --rebase your branch as I've made some changes and merged origin/master on it. As Sven describes, the following can be added:
- Support to specify keyboard
- Support to supply userdata (this would be sent as base64 encoded string as part of the api request)
- The add to instance group (existing groups may be listed using the listInstanceGroups API)

Rest is largely done, but I'm still reviewing/testing it.

utchoang:
@rhtyd cc @svenvogel 
As from https://github.com/apache/cloudstack-primate/pull/7, the template/iso is filtered by osType, which causes incorrect pagination. To solve it, I intend to change the filter from osType to filter/show by featured/community/shared etc. Please let me know what you think.
image

rhtyd:
@utchoang is this ready for review/testing now?

utchoang:
@rhtyd I completed fixes. Please review again

utchoang:
The requirements for this task are:
* Keep the wizard in a pop-over
* Add support for both basic and advanced, ideally we would prefer the similar kind of layout/steps like in the old UI's wizard

I think the new UI already meets these demands:
1. The wizard is a popup
image

2. There's support info for both Basic and Advanced when it is selected. The layout is in form of steps.
image

Could you please explain what I should change?

shwstppr:
@utchoang we don't have the functionality to add a new network on the fly like in old UI?
Screenshot from 2020-03-25 18-17-10
And can we make changes to pre-select(first entry from the list) things like zone, compute offering, etc.

Deployment works with minimal changes to variables. Will try more variations and will try to provide some fixes tomorrow. cc @rhtyd

rhtyd:
There are still some issue but largely the feature works for some combination, I'll merge and hopefully the bugs can be fixed over time cc @shwstppr @utchoang @vladimirpetrov

rhtyd:
Added bug https://github.com/apache/cloudstack-primate/issues/254

PR #149, opened 2020-02-05
Add tier select + update to use <a-table>

RitchieVincent:


rhtyd:
@RitchieVincent lgtm, tested it. Please add pagination after a-table like the other tabs/view you did on the vpc component.

PR #228, opened 2020-03-17
Missing functionality - Templates > ISO - change order

utchoang:
@rhtyd cc @svenvogel 
I've completed fix change order template > Iso. Please review and let me know your comments.
image


rhtyd:
Not necessary; the feature is not supported for old UI cc @vladimirpetrov

PR #285, opened 2020-04-06
view: fix for update template iso permissions form

### LINKED ISSUE #278: [BUG] ISO - Update permissions form does not autodiscover operation, accounts and projects
**Describe the bug**
Update permissions form does not autodiscover operation, accounts and projects

**To Reproduce**
1. Go to ISOs
2. Select an iso
3. Click on Update permissioins
4. See that fields are not autodiscovering operation/account/project

**Expected behavior**
There should be dropdown option for the fields 

**Screenshots**
Screenshot 2020-04-02 at 16 38 52

shwstppr:
Fixes #278 

Screenshot from 2020-04-06 14-54-07


rhtyd:
@shwstppr move the copy and delete buttons/action to the iso and template zones components.

PR #289, opened 2020-04-10
view: fix list group action

### LINKED ISSUE #162: [BUG] Group action buttons don't work properly (Events, Alerts, VM)
**Describe the bug**
The delete/archive event button has very funny behavior. If no check box is selected it still deletes/arhives the top item, which is not very obvious. The selection boxes seem to be ignored, but still sometimes they are deselected after archiving (deleting) of event.

Notes by @rhtyd 
Group action buttons need to be supported for vm, event and alert views.

**To Reproduce**
1. Go to 'Events'.
2. Click on the 'Delete' or 'Archive' event icon.
3. Inspect the result.
4. Mark some (or all) of the events (check the selection boxes), then click again on the 'Delete' or 'Archive' event icon.

Result: The top event is deleted (archived), regardless of the checked boxes and sometimes they are deselected.

**Expected behavior**
Exactly and only the marked events must be deleted/archived.

**Screenshots**
None.

 - OS: Windows 10 Pro x64
 - Browser: Chrome 
 - Version: 80.0.3987.106 (Official Build) (64-bit)

**Additional context**
None.

shwstppr:
Fixes #162


src/views/AutogenView.vue
```diff
@@ -726,6 +741,11 @@ export default {
           var hasJobId = false
           api(this.currentAction.api, params).then(json => {
             // set action data for reload tree-view
+            if (this.currentAction.api.startsWith('delete') ||
```
rhtyd, line 744: I would prefer that we simply run the api of the action against all the selected items; the only exception could be if the parameter the API accepts is `ids` then instead of looping+calling on all items; we can simply call a single API. Let's discuss.

shwstppr: @rhtyd  made the fix for looping+calling API for all selected items when `ids` parameter is not available.
Also, when `groupAction` is set to `true` and there is selection in list then we ignore show property of the action.

shwstppr:
@rhtyd refactored some code but couldn't figure how to clear selection for APIs that remove items.
Either we clear selection for all API or the current method

rhtyd:
@blueorangutan hi

rhtyd:
config -> method/handle to process list of selected items -> return list of params to be called against currentAction.api

rhtyd:
Fixed in 750ea60696aefa0b33913d9e195f37b51f24ac4c

PR #290, opened 2020-04-10
view: notify admin to restart MS for non-dynamic settings

### LINKED ISSUE #83: Detect and only notify admin to restart mgmt server depending if setting needs it
Refer:
https://github.com/apache/cloudstack/pull/3729

shwstppr:
Fixes #83

### DIFF
--- a/src/components/view/ListView.vue
+++ b/src/components/view/ListView.vue
@@ -266,14 +266,19 @@ export default {
       api('updateConfiguration', {
         name: record.name,
         value: this.editableValue
-      }).then(() => {
+      }).then(json => {
         this.editableValueKey = null
 
         this.$message.success('Setting Updated: ' + record.name)
-        this.$notification.warning({
-          message: 'Status',
-          description: 'Please restart your management server(s) for your new settings to take effect.'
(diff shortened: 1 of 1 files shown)


src/components/view/ListView.vue
```diff
@@ -266,14 +266,19 @@ export default {
...
-      }).then(() => {
+      }).then(json => {
         this.editableValueKey = null
 
         this.$message.success('Setting Updated: ' + record.name)
-        this.$notification.warning({
-          message: 'Status',
-          description: 'Please restart your management server(s) for your new settings to take effect.'
-        })
+        if (json.updateconfigurationresponse &&
```
rhtyd, line 273: @shwstppr would this work if API doesnot have the isdynamic in the API response?

shwstppr: @rhtyd should be okay with `!json.updateconfigurationresponse.configuration.isdynamic &&`

https://jsfiddle.net/shwstppr/x5undf7a/

rhtyd: okay

PR #296, opened 2020-04-17
view: fix api error handling

### LINKED ISSUE #293: [BUG] Infrastructure > System VMs > Run Diagnostics does not show the command output
**Describe the bug**
The action 'Run Diagnostics' from the System VMs screen is not showing the result of the command.

**To Reproduce**
1. Go to 'Infrastructure > System VMs' and click on a system VM
2. Click on the 'Run Diagnostics' button, then choose 'ping' operation with some argument - '8.8.8.8' for example.

**Expected behavior**
The result of the command must be shown, as in the old UI. Currently only a 'Network error' is shown instead and it's not clear where exactly is the problem.

**Screenshots**
image

image

 - OS: Ubuntu 18.04 LTS
 - Browser: Chrome
 - Version: 81.0.4044.92 (Official Build) (64-bit)

**Additional context**
I'm not sure if this functionality even works.

### LINKED ISSUE #294: [BUG] Can't add load balancer
**Describe the bug**
When trying to add a new load balancer, the VM selection popup doesn't load.
It's impossible to add a load balancer via the UI because of this.

button, 
**To Reproduce**
1. Go to an allocated IP address
2. Select the "Load Balancing" tab
3. Enter values into the Name, Public Port, Private Port, Algorithm, Protocol fields
4. Click on the "Add" button

**Expected behavior**
The VM selection dialogue should pop up.

**Screenshots**
Screenshot_20200416_144147

 - OS: Linux
 - Browser: Firefox
 - Version: 74.0.1

**Additional context**
The issue seems to happen when the listVirtualMachines call sent by the "Add VMs" handler returns an empty list, like this:
```json
{"listvirtualmachinesresponse":{}}
```

The console log prints the message:
```
TypeError: t.response is undefined           LoadBalancing.vue:935:10
```

This is with the latest HEAD (commit bcf86f450d6aede6bad5d556a1cf81fe00ccfd2c )

shwstppr:
Fixes #294 
Screenshot from 2020-04-17 17-09-54

Fixes parsing invalid error object


shwstppr:
@rhtyd I've fixed that message comparison issue. I don't have a better solution right now for the whole notification thing.
Maybe we can add a parameter for `api` function call such as `notifyError` (with false as the default value) and then remove `this.$notification.error(...` code for API calls from the whole codebase where we are just simply showing error in notifications.
Although I feel better to not do it in TP milestone.

onitake:
I tested b3805f8 and it solved the issue reported in #294 . I haven't tested 944c5f6 yet.
Is there a specific reason why you changed `(x && y) || z` to `x ? y : z`?
A general dislike of the ternary operator notwithstanding, I think the ?: construct is easier to read - but YMMV.

shwstppr:
@onitake change was due to an internal review comment. I like `?:` myself, easier to understand.
Although, we might have to refactor the error handling more exhaustively as mentioned in previous comment.

rhtyd:
discuss - come up with a general purpose way of declaring errors?

shwstppr:
@rhtyd I've added global method and refactored some code (some more to be done).
Does this look okay or should we handle error notification in `src/api/index.js` itself with some flag in `api` method?

src/utils/plugins.js
```diff
@@ -77,3 +78,35 @@ export const pollJobPlugin = {
...
+        if (desc === '' && error.response.data) {
+          const responseKey = _.findKey(error.response.data, 'errortext')
+          if (responseKey) {
+            desc = error.response.data[responseKey].errortext
+          }
+        }
+      }
+      notification.error({
+        message: msg,
+        description: error && error.response && 'x-description' in error.response.headers ? error.response.headers['x-description'] : error.message,
```
shwstppr, line 107: @rhtyd shoud this read, `description: desc`?

PR #309, opened 2020-04-24
added color changes for different progress value

### LINKED ISSUE #308: [BUG] Apply threshold logic to colour pie-progress red if threshold exceeds 80%
On the dashboard, check and enforce that progress bar is shown red if value exceeds 80% (see code in src/view/dashboard/). This is for the admin dasboard.

cc @shwstppr

shwstppr:
Fixes #308 

```
value >= 80 : 'red'
default: '#1890ff'
```

Default colour value from here, https://github.com/apache/cloudstack-primate/blob/master/vue.config.js#L102

### DIFF
--- a/src/views/dashboard/CapacityDashboard.vue
+++ b/src/views/dashboard/CapacityDashboard.vue
@@ -57,9 +57,10 @@
               <h4>{{ $t(stat.name) }}</h4>
               <a-progress
                 type="dashboard"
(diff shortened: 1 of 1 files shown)


rhtyd:
@shwstppr can you add screenshot for all the thresholds? I think we don't want any different/non-default colour for thresholds below 75/80 (idea is to have feature parity with old UI).

PR #297, opened 2020-04-17
view: fix runDiagnostic output display

shwstppr:
Fixes: #293 

Screenshot from 2020-04-17 16-04-21

Issue #202, opened 2020-03-06
[BUG] Missing functionality - Infrastructure > Virtual routers > Get diagnostic data


vladimirpetrov:
**Describe the bug**
The action 'Get diagnostic data' from the Virtual routers screen is missing in Primate UI.

**To Reproduce**
1. Go to 'Infrastructure > Virtual routers' and click on a virtual router
2. Look at the possible actions - there's no 'Get diagnostic data'.

**Expected behavior**
There should be a button 'Get diagnostic data' as in the old UI.

**Screenshots**
![image](https://user-images.githubusercontent.com/12384665/76098842-df290600-5fd2-11ea-90fc-bf4886267bb1.png)

**Desktop (please complete the following information):**
 - OS: Windows 10 Pro x64
 - Browser: Chrome
 - Version: 80.0.3987.132 (Official Build) (64-bit)

**Additional context**
Probably related to #201

borisstoyanov:
@rhtyd What is the cloudstack version we're aiming to be feature parity with?

rhtyd:
@borisstoyanov we started with 4.13, we'll try and see if we can support 4.14 as well.

rhtyd:
@vladimirpetrov @borisstoyanov please test and close against master, thanks.

vladimirpetrov:
There is a button now and it says it completes the task successfully but does not show the URL to download the diagnostic data:

![image](https://user-images.githubusercontent.com/12384665/79842242-31768880-83c1-11ea-9d74-b1278e6ba1f6.png)

like it did in the old UI, for example:

![image](https://user-images.githubusercontent.com/12384665/79842319-51a64780-83c1-11ea-8cc8-ea9c70dbc1e0.png)

rhtyd:
@shwstppr please see ^^

vladimirpetrov:
Fix confirmed, closing the issue.

Issue #201, opened 2020-03-06
[BUG] Missing functionality - Infrastructure > System VMs > Get diagnostic data


vladimirpetrov:
**Describe the bug**
The action 'Get diagnostic data' from the System VMs screen is missing in Primate UI.

**To Reproduce**
1. Go to 'Infrastructure > System VMs' and click on a system VM
2. Look at the possible actions - there's no 'Get diagnostic data'.

**Expected behavior**
There should be a button 'Get diagnostic data' as in the old UI.

**Screenshots**
![image](https://user-images.githubusercontent.com/12384665/76098145-9ae93600-5fd1-11ea-8d75-371823458e81.png)

**Desktop (please complete the following information):**
 - OS: Windows 10 Pro x64
 - Browser: Chrome
 - Version: 80.0.3987.132 (Official Build) (64-bit)

**Additional context**
Probably related to #202

rhtyd:
@vladimirpetrov @borisstoyanov please test and close against master, thanks.

vladimirpetrov:
There is a button now and it says it completes the task successfully but does not show the URL to download the diagnostic data.

![image](https://user-images.githubusercontent.com/12384665/79841881-b6ad6d80-83c0-11ea-9828-27eac25df4a1.png)

as it did in the old UI, for example:

![image](https://user-images.githubusercontent.com/12384665/79841835-aa291500-83c0-11ea-9e55-c3d922919477.png)

rhtyd:
@shwstppr please see ^^ did you use a `response` param?

shwstppr:
@rhtyd please have look on new PR when possible

rhtyd:
@vladimirpetrov @borisstoyanov PR merged, please re-test against master. This will require a real env (say kvm etc), we may get unexpected results for simulator.

vladimirpetrov:
Fix confirmed, closing the issue.

PR #312, opened 2020-04-24
view: fix getDiagnostics response display

shwstppr:
Fixes response display for,
#201 
#202 
Screenshot from 2020-04-24 16-07-48
Screenshot from 2020-04-24 16-07-21

Notification view label will be fixed with #297


src/config/section/infra/routers.js
```diff
@@ -102,7 +102,8 @@ export default {
         targetid: {
           value: (record) => { return record.id }
         }
-      }
+      },
+      response: (result) => { return ((result && result.diagnostics && result && result.diagnostics.url && `Please click the link to download the retrieved diagnostics: <p><a href='${result.diagnostics.url}'>${result.diagnostics.url}</a></p>`) || 'Invalid response') }
```
shwstppr, line 106: @rhtyd can we perform the translation in JS files?
I've to use string key from here, get the translated string and then replace URL. ELse I'll have to add code in Autogen to explicitly address this API response

PR #231, opened 2020-03-17
storage: fix volume actions

shwstppr:
Fixes #225

PR #313, opened 2020-04-24
Bootintobios

DaanHoogland:
UI implementation for https://github.com/apache/cloudstack/pull/4021


DaanHoogland:
@rhtyd @shwstppr can you review please?

src/config/section/compute.js
```diff
@@ -86,7 +86,7 @@ export default {
           dataView: true,
           groupAction: true,
           show: (record) => { return ['Stopped'].includes(record.state) },
-          args: ['podid', 'clusterid', 'hostid'],
+          args: ['podid', 'clusterid', 'hostid', 'bootintobios'],
```
rhtyd, line 89: @DaanHoogland is this feature allowed for users as well?

src/views/compute/DeployVM.vue
```diff
@@ -275,6 +275,16 @@
                   </div>
                 </template>
               </a-step>
+              <a-step
+                :title="this.$t('enter BIOS setup')"
```
rhtyd, line 279: `this.$t..` can be written as `$t(...). Instead of a string, can you use a key, we usually encourage the same key as the API arg/parameter if applicable. You can then define the key and the language translation in src/locales/en.json

src/views/compute/DeployVM.vue
```diff
@@ -275,6 +275,16 @@
...
+                :status="zoneSelected ? 'process' : 'wait'">
+                <template slot="description">
+                  <div v-if="zoneSelected">
+                    <a-checkbox
+                      @change="enterBiosChoiceChanged">enter BIOS setup</a-checkbox>
```
rhtyd, line 284: if it's just a setter, we can also write shorthand like this:
```
@change="val => bootIntoBios = val"
```

src/views/compute/DeployVM.vue
```diff
@@ -918,6 +933,8 @@ export default {
         deployVmData.keypair = values.keypair
         deployVmData.name = values.name
         deployVmData.displayname = values.name
+        // step 8: enter setup
+        deployVmData.bootintobios = this.bootintobios
```
rhtyd, line 937: I think this setting can be at the top step itself where we're getting name, keyboard, etc.

src/config/section/compute.js
DaanHoogland: yes as i understand it it is

src/views/compute/DeployVM.vue
DaanHoogland: i'll test this, i didn't get any other way to work, yet.

src/views/compute/DeployVM.vue
DaanHoogland: ok, kinda makes sense

src/views/compute/DeployVM.vue
shwstppr, line 284:
```suggestion
                      @change="event => { this.bootintobios = event.target.checked }">enter BIOS setup</a-checkbox>
```
@DaanHoogland

DaanHoogland:
thanks guys, I'll address your feedback, soon.

src/config/section/compute.js
DaanHoogland: last args should be conditional on `hypervisor === vmware`. will be done after #231 has been merged with its [AutoGenView.vue changes](https://github.com/apache/cloudstack-primate/pull/231/files#diff-83f2f8976f46b2444d31fcd1895d8165R548)

src/config/section/compute.js
```diff
@@ -104,7 +104,8 @@ export default {
           icon: 'reload',
           label: 'label.action.reboot.instance',
           dataView: true,
-          show: (record) => { return ['Running'].includes(record.state) }
+          show: (record) => { return ['Running'].includes(record.state) },
+          args: ['bootintobios']
```
DaanHoogland: this args field should be conditional on `hypervisor === vmware`. will be done after #231 has been merged with its [AutoGenView.vue changes](https://github.com/apache/cloudstack-primate/pull/231/files#diff-83f2f8976f46b2444d31fcd1895d8165R548)

DaanHoogland:
thanks @rhtyd and @shwstppr . I addressed your comments, please have a look. And please note that this code will at best do nothing until [cloudstack 4021](https://github.com/apache/cloudstack/pull/4021) is merged.

src/config/section/compute.js
```diff
@@ -104,7 +104,8 @@ export default {
           icon: 'reload',
           label: 'label.action.reboot.instance',
           dataView: true,
-          show: (record) => { return ['Running'].includes(record.state) }
+          show: (record) => { return ['Running'].includes(record.state) },
+          args: ['bootintosetup']
```
DaanHoogland, line 108: @shwstppr , I should be able to replace this with
```
args: (record) -> { return record.hypervisorType === 'vmware' ? ['bootintosetup'] : [] }
```
am I? It seems not to work. How can I find what to use as parameter of the anonymous function and what attributes it has?

tnx

shwstppr, line 108:
```suggestion
          args: (record) => { return record.hypervisor === 'VMware' ? ['bootintosetup'] : [] }
```
@DaanHoogland 
This should work but currently it seems record is coming undefined from AutogenView. I've created PR #327 to fix that.
`record` here will be the response for `listvirtualmachinesmetrics` response item

DaanHoogland: ah to bad, I want to know the hypervisor type for the machine. @shwstppr any tips?

shwstppr: @DaanHoogland `record.hypervisor` with give you that.

VMs in primate are listed with listvirtualmachinesmetrics API and it only has hypervisor param. No parameter named hypervisortype in API response.

DaanHoogland: hypervisor contains the type, tnx. There is something confusing in the API doc, it only lists it as a request parameter. I think I should be fine.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@DaanHoogland please ping me when this is ready for review

DaanHoogland:
@rhtyd it is ready for review, but the cloudstack code is not merged yet.

rhtyd:
@DaanHoogland can you fix the merge conflicts? I think we merge this in Primate as soon as we've any test confirmation. Thanks.

DaanHoogland:
@shwstppr @davidjumani @rhtyd I rebased but my en.json file does not seem ok. to much diff. Can you check if I did the reverse of what was intended?

rhtyd:
@DaanHoogland not urgent - the next time you get chance to work on this - can you revert your changes to en.json; keep the version from upstream and then sort/diff what you need to specific keys for your feature.

DaanHoogland:
@rhtyd fixed my booboo (hope the late mister Hawkins approves)

rhtyd:
@blueorangutan package

rhtyd:
@borisstoyanov @vladimirpetrov have you also tested this Primate PR against the upstream PR (merged upstream now) https://github.com/apache/cloudstack/pull/4021?

borisstoyanov:
@rhtyd yes, primate changes has been covered also

src/config/section/compute.js
```diff
@@ -89,7 +89,18 @@ export default {
...
           groupAction: true,
           show: (record) => { return ['Stopped'].includes(record.state) },
-          args: (record, store) => { return ['Admin'].includes(store.userInfo.roletype) ? ['podid', 'clusterid', 'hostid'] : [] },
+          args: (record, store) => {
+            var fields = []
+            if (store.userInfo.roletype === 'Admin') {
+              fields = ['podid', 'clusterid', 'hostid']
+            }
+            if (record.hypervisor === 'VMware') {
+              if (store.apis.startVirtualMachine.params.filter(x => x.name === 'bootintosetup').length > 0) {
```
rhtyd, line 98: @DaanHoogland I've tested and fixed a case here - we want to make Primate work with all recent/pass/future version of CloudStack; so if there's a feature adding new params for autogenerated forms, we can check if the param is available by checking the API discovery /cc @shwstppr @davidjumani

PR #316, opened 2020-04-28
view: list filtering

### LINKED ISSUE #213: [BUG] Missing Instance/VM 'Filter by' functionality
**Describe the bug**
It's impossible to filter by owned, running, stopped or destroyed instances, as in the old UI.

**To Reproduce**
1. Go to 'Compute > Instances' - the 'Filter by' functionality is missing.

**Expected behavior**
There should be a way to filter by owned, running, stopped or destroyed instances, as in the old UI.

**Screenshots**
image

 - OS: Windows 10 Pro x64
 - Browser: Chrome
 - Version: 80.0.3987.132 (Official Build) (64-bit)

**Additional context**
This filtering is missing also in the following pages:
- Templates

shwstppr:
Fixes #213

### DIFF
--- a/src/config/router.js
+++ b/src/config/router.js
@@ -64,7 +64,8 @@ export function generateRouterMap (section) {
           related: child.related,
           actions: child.actions,
           treeView: child.treeView ? child.treeView : false,
-          tabs: child.treeView ? child.tabs : {}
+          tabs: child.treeView ? child.tabs : {},
+          filter: child.filter
         },
         component: component,
         hideChildrenInMenu: true,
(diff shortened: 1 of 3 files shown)


shwstppr:
@rhtyd will something like this make sense?

rhtyd:
@shwstppr see table/column filtering in antd, if that is not possible we can go with your approach

shwstppr:
@rhtyd I've implemented table column filtering with antd here https://github.com/shapeblue/cloudstack-primate/tree/feature-list-filter-alt
Screenshot from 2020-05-04 12-14-03

Since default filtering UI shows checkboxes and API won't support multiple states for filtering, I'll have to implement custom UI

rhtyd:
@blueorangutan package

rhtyd:
Implemented that with a simpler fix, closing. Thanks.

PR #317, opened 2020-04-29
view: fix role-base checks on components to disable if role is not allowed an operation

### LINKED ISSUE #310: [BUG] Add role based access checks for custom UI elements/forms
Log in as read-only root admin and you'll still see some buttons tabs that must be hidden from the logged in user. For this in all such custom components (mostly tabs), a v-if check if needed to show/hide elements based on `apiName in $store.getters.api...` check.

shwstppr:
Fixes #310


shwstppr:
@rhtyd I've tried to address all possible cases, please check

shwstppr:
@rhtyd I've verified and made changes to disable buttons/actions across UI. I've not changed it in custom forms which are called for certain API and will be taken care of by permissionchecker

rhtyd:
@shwstppr  can you fix the conflict? Thanks.

rhtyd:
@blueorangutan package

PR #330, opened 2020-05-07
Add unmanage VM button

nvazquez:
Add the Unmanage VM button on the instance view for VMware running VMs:

image


rhtyd:
@nvazquez is this ready for review?

src/config/section/compute.js
```diff
@@ -319,6 +319,20 @@ export default {
...
+          api: 'unmanageVirtualMachine',
+          icon: 'disconnect',
+          label: 'label.action.unmanage.virtualmachine',
+          args: ['id'],
+          dataView: true,
+          groupAction: true,
+          show: (record) => { return ['Running', 'Stopped'].includes(record.state) && record.hypervisor === 'VMware' },
+          mapping: {
+            id: {
+              value: (record, params) => { return record.id }
```
rhtyd, line 332: we use mapping when the key is not the same as specified in the `args`, so this will still work without adding the mapping here

rhtyd:
@blueorangutan package

PR #331, opened 2020-05-11
view: cks cluster async actions fix

shwstppr:
- Poll cluster async actions
- Fixes related to scale cluster
- Action labels from translations


shwstppr:
@rhtyd please see if this can be included in TP

PR #343, opened 2020-05-16
 Allow configuring IPv6 default guest network when deploying a new Zone

GabrielBrascher:
Advanced Network with Security Groups is a great option for deploying a Zone with IPv6 address support. However, it is not possible to set IPv6 DNS + IPv6 CIDR + IPv6 gateway via the zoneWizard UI.

This PR adds a small enhancement on the UI Zone deployment to allow deploying a Zone with Advanced Network with Security Groups + IPv6 CIDR, IPv6 gateway, and IPv6 DNS.

Note that API [1] offers full support for such action, therefore there is no need for changing the API. It is just a small enhancement on UI that might be useful in the meantime that we wait for the new ACS UI.

[1] https://cloudstack.apache.org/api/apidocs-4.13/apis/createNetwork.html
image

### DIFF
--- a/src/views/infra/zone/ZoneWizardZoneDetailsStep.vue
+++ b/src/views/infra/zone/ZoneWizardZoneDetailsStep.vue
@@ -81,7 +81,7 @@
       <a-form-item
         :label="$t('label.ipv6.dns1')"
         v-bind="formItemLayout"
-        v-if="isAdvancedZone && !securityGroupsEnabled"
+        v-if="isAdvancedZone"
         has-feedback>
         <a-input
           v-decorator="['ipv6Dns1', {
@@ -102,7 +102,7 @@
       <a-form-item
         :label="$t('label.ipv6.dns2')"
         v-bind="formItemLayout"
-        v-if="isAdvancedZone && !securityGroupsEnabled"
+        v-if="isAdvancedZone"
         has-feedback>
         <a-input
           v-decorator="['ipv6Dns2', {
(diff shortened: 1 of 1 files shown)


rhtyd:
Hi @GabrielBrascher - have you tested this by deploying a zone, does it pass/fail? Thanks.
@blueorangutan package

GabrielBrascher:
@rhtyd I will run a few more tests this week.

src/views/infra/zone/ZoneWizardZoneDetailsStep.vue
```diff
@@ -120,6 +120,43 @@
           }]"
         />
       </a-form-item>
+      <a-form-item
+        :label="$t('ip6cidr')"
+        v-bind="formItemLayout"
+        v-if="isAdvancedZone && securityGroupsEnabled"
```
rhtyd, line 126: Is this setting/field only taken into consideration when SG is enabled @GabrielBrascher ?

src/views/infra/zone/ZoneWizardZoneDetailsStep.vue
```diff
@@ -120,6 +120,43 @@
...
+                initialValue: ip6cidr
+              }
+            ]
+          }]"
+        />
+      </a-form-item>
+            <a-form-item
+              :label="$t('ip6gateway')"
+              v-bind="formItemLayout"
+              v-if="isAdvancedZone && securityGroupsEnabled"
```
rhtyd, line 142: Is this setting/field only taken into consideration when SG is enabled @GabrielBrascher ?

rhtyd:
@blueorangutan package

rhtyd:
Tested on live env, LGTM; fixed translation label issue.

src/views/infra/zone/ZoneWizardZoneDetailsStep.vue
GabrielBrascher: Sorry for the late reply @rhtyd.

So far there is no support for IPv6 via VPC/VRs. Therefore, the zone must be an Advanced zone with Security Group enabled to handle IPv6 allocation and ACLs.

Thanks for the review!

src/views/infra/zone/ZoneWizardZoneDetailsStep.vue
GabrielBrascher: IPv6 gateway and CIDR are handled on an advanced network with security group enabled. I don't see any harm in adding it to any type of zone network; however, that might lead users to believe that ACS will provide IPv6 support to the zone.

rhtyd:
Thanks @GabrielBrascher for your replies, the PR was already tested and merged.

Issue #339, opened 2020-05-14
[FEATURE] Boot type and boot mode dropdowns in the deploy wizzard


DaanHoogland:
**Is your feature request related to a problem? Please describe.**
In 4.14 the ability to deploy with a specified boot type and boot mode, i.e. UEFI/secure, is possible. This should be reflected in the primate UI.
 
**Describe the solution you'd like**
extra dropdowns with the possible options for boot type and boot mode should be available in the overview (top level) step of the deploy wizard.

**Describe alternatives you've considered**
Other places could be with the template and hardware dimnesions choices but these does not seem better locations.

**Additional context**
Boot type can be 'BIOS' or 'UEFI' and mode can be 'legacy' or 'secure'.

rhtyd:
@utchoang can you pick this up next. Please check the old uI (http://primate-qa.cloudstack.cloud:8080/client/) to see the boot type/mode options in the VM deployment wizard - we want that for Primate's vm deployment wizard as well.

Thanks.

PR #364, opened 2020-05-27
Add boottype and bootmode to VM instance

utchoang:
Open for https://github.com/apache/cloudstack-primate/issues/339

@rhtyd cc @svenvogel I've completed this task. Please review and let me know your comments
image


src/views/compute/DeployVM.vue
```diff
@@ -80,6 +80,25 @@
...
+                    </a-form-item>
+                    <a-form-item :label="this.$t('label.vm.bootmode')">
+                      <a-select
+                        v-decorator="['bootmode']"
+                      >
+                        <a-select-option v-for="bootMode in options.bootModes" :key="bootMode.id">
+                          {{ bootMode.description }}
+                        </a-select-option>
+                      </a-select>
+                    </a-form-item>
```
DaanHoogland, line 101: This needs to be for vmware and kvm only i think

utchoang: @rhtyd Do you think about this? Because on https://github.com/apache/cloudstack/blob/d4b537efa7ed27ef88b735a22493808b61529186/ui/scripts/ui-custom/instanceWizard.js. I don't see any description for show/hidden of boot type & bootmode

DaanHoogland: @utchoang I think that is an oversight or short cut in the new UI. The UI is being phased out and the feature rather new. I think it slipped attention.

src/views/compute/DeployVM.vue
```diff
@@ -80,6 +80,29 @@
                         :options="keyboardSelectOptions"
                       ></a-select>
                     </a-form-item>
+                    <a-form-item
```
rhtyd, line 83: @utchoang cc @DaanHoogland - the hypervisor would probably only be available once (a) the template is selected or (b) ISO is selected and hypervisor is selected; should this be moved after the template section then?

utchoang: @rhtyd I moved it below the template/iso section

rhtyd:
@blueorangutan package

rhtyd:
Headsup @utchoang - I've moved the name/details to the end step because of dependencies from later steps

Issue #359, opened 2020-05-26
[FEATURE] Add support for UI customisation


rhtyd:
- [ ] Add supporton branding the UI: logos, and theme (refer: https://www.antdv.com/docs/vue/customize-theme/)
- [ ]  Add user documentation on branding
- TBD - Introduce/use a config.js or config.json file for high level custom options (such as keyboards, configs etc.)

rhtyd:
@utchoang cc @svenvogel can you have a look at the issue of adding a theme for Primate. Ideally a json file that (1) Primate loads and applies, (2) the json/theme file does not get packed/bundled with rest of the code and is available say similar to index.html on the root `dist` directory. Same for any other branding logos/images.

utchoang:
@rhtyd cc @svenvogel  In primate, the logo is a component. But under this mission, I will change it to custom, right?

rhtyd:
I think what we can do is to move the logo (png and svgs) such that users can simply replace the files with their chose png/svg files after they use the dist/archive/deb/rpm used to install Primate

rhtyd:
Related legacy UI for tracking - https://github.com/apache/cloudstack/pull/3258

rhtyd:
Got a PR

PR #372, opened 2020-06-02
Add support for UI customisation

utchoang:
Open for https://github.com/apache/cloudstack-primate/issues/359

* [x] Add support on branding the UI: logos, and theme
* [x] Add user documentation on branding


rhtyd:
@utchoang just a note - when you test please do a `npm run build` and copy/run the dist files either under nginx/docker or within a CloudStack management server. The idea is end-user would install rpm/deb/docker container etc and should be able to make changes to the theme.json (or similar) config file.

svenvogel:
@utchoang this is a important thing. Cool! @rhtyd you mean they will fetched in the build process or later?

rhtyd:
@svenvogel @utchoang  the expectation as a user would be that after one installs Primate (say using the deb/rpm/archive or even docker image) then one should be able to edit the config.json to customise the theme/colours. Therefore in order to test, we need to build it (npm run build) and from the `dist` test by either copying to a mgmt server env or use docker/nginx reverse proxy (see README on how to setup nginx reverse proxy with the build Primate dist), and then change colours in theme.json (or any other config file/location) and refresh UI to see theme colours changed.

It may be necessary to load less.css etc to make it work @utchoang - I've not explore that myself, but as @svenvogel mentions it would be really cool feature to have!

utchoang:
@rhtyd I have researched and found out that I have understood how it works and how to do it. I will finish it as soon as possible. Thanks.

rhtyd:
Fantastic @utchoang - no problem, no one's in a hurry, we're just excited about this feature.

utchoang:
@rhtyd cc @svenvogel I tried to customize the theme by apply https://www.npmjs.com/package/antd-theme-webpack-plugin. I was able to change the text color, button color, etc. But for other properties such as font size, it only applies after rebuilding again.
Can you have any solution for this?

rhtyd:
No, I think wr don't need to configure fonts, sizes etc. The feature of providing some way to customise the theme to the end user by means of some config file is enough.
@utchoang so does it work in the built dist and changing just the config.json?

utchoang:
@rhtyd Yes. Just change the information in the config.json file. Then refresh the page and it will be applied.

rhtyd:
Fantastic, well done @utchoang - is it ready for review now, or you still doing changes?

utchoang:
@rhtyd I need to edit a few more points. Also, do I need to write user documentation?

rhtyd:
A short markdown documentation in `docs/` would be great @utchoang thanks

svenvogel:
@utchoang yes a doc would be good

svenvogel:
@utchoang we need to change the logo in the config.json?

rhtyd:
@blueorangutan package

utchoang:
@svenvogel Yes. I and Rohit have exchanged about this on https://github.com/apache/cloudstack-primate/issues/359#issuecomment-638573176

svenvogel:
> No, I think wr don't need to configure fonts, sizes etc. The feature of providing some way to customise the theme to the end user by means of some config file is enough.
@utchoang so does it work in the built dist and changing just the config.json?

@rhtyd we need to change the size of the logo too. We need to find a good way to use this with config file? I would prefer to add something or how we could change this by an reproducable way? cc @utchoang

rhtyd:
@svenvogel which logo - the login logo, or the left-top side logo that shows after we login? Or any other? I think instead of config file, we can have a folder to put logos (svg/png, preferably svgs) and customisation can be done by replacing the logo file than change in config.json file, what do you think?

public/config.json
```diff
@@ -0,0 +1,15 @@
+{
+  "theme": {
+    "@primary-color": "#1890ff",
```
rhtyd, line 3: @utchoang are these the complete list of variables, this has more variables - https://www.antdv.com/docs/vue/customize-theme/#Ant-Design-Vue-Less-variables ? Thanks (excluding fonts is okay I think). cc @svenvogel ?

utchoang: @rhtyd I checked that there are some attributes like @ link-color that don't seem to be used in the project so I deleted them. So you will notice that there are some incomplete attributes.

rhtyd:
@blueorangutan package

rhtyd:
@utchoang can you move the following keys to the config.json from the `src/config/settings.js`; which would allow the admin configure the API request path, the doc base URL path and the app/tab title prefix:
```
  apiBase: '/client/api',
  docBase: 'http://docs.cloudstack.apache.org/en/latest',
  appTitle: 'CloudStack',

```

utchoang:
@rhtyd cc @svenvogel 
About changing the logo size. I think the @Seven ideas would be good because the end-user only needs to change the logo size in the config file instead of using the third software to adjust it. I will try to find a way to interfere with the Ant Design style and I hope it works.

rhtyd:
@utchoang do you know which logo is @svenvogel talking about the logo in favicon, or login screen, or post-login (on left-top side) or anything else?

utchoang:
@rhtyd I think all logo (login screen, left-top side,...)

utchoang:
@rhtyd He means when we wanted to change another logo, maybe its size would not be right, so just adjust it using the config file instead of the third software.

rhtyd:
@utchoang okay, I guess first thing then - move all the banner and logo svgs to public/assets (I think you've moved logo, maybe even the error/403/404/500 pngs). I think the size is bound by the navigation bar width - that would be a limitation. Generally, like legacy UI, all images have a fixed sized - otherwise it will affect how the UI looks (it will cause other components/parts to spill out).

rhtyd:
Note for myself - we'll need to check and get some of the config from legacy UI such as keyboard options: https://github.com/apache/cloudstack/pull/3258/files

svenvogel:
@rhtyd @utchoang yes maybe the layout will give us only a fit size of the icon/logs and i agree with a copy and paste of the logo in a folder and overwrite them into. the cool thing from our templater was in the old UI there was a possibility to overwrite things with a "config" file if anyone want change different things in the UI before we build them. we dont need to hack in the files directly e.g. he want to change the size from logo (i know this will maybe no fit into the layout/design) or change other things.

rhtyd:
@utchoang let me know when you're done with any other changes, I'll help review and merge this

utchoang:
@rhtyd I will notify you when it is completed with the new user document.

rhtyd:
@utchoang please resolve conflict and ping me if/when ready for review; thanks.

utchoang:
@rhtyd It ready for review. Thanks.

rhtyd:
@blueorangutan package

rhtyd:
@utchoang can you check the build - http://primate-qa.cloudstack.cloud:8080/client/pr/372/#/dashboard
- The default logo (after login, on left-top side) is missing, and the login banner
- There's a fetching configuration message which is under loading (something went wrong?)
- Maybe the issue is assuming the path/url of the config.json (could we try the local path like ./ instead of /)

Thanks for the effort, looks great so far

utchoang:
@rhtyd I think to turn to `. /config.json` did not solve the problem. Because the URL for the fetch configuration file is wrong. Instead, I think using `process.env.BASE_URL` to fetch the config.json file is better.
What do you think about this?
image

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

PR #383, opened 2020-06-05
Fixing wrong labels

davidjumani:
Fixes incorrect labels as well as add new ones to `en.json`


src/locales/en.json
```diff
@@ -704,7 +706,6 @@
 "label.dns1": "DNS 1",
 "label.dns2": "DNS 2",
 "label.domain": "Domain",
-"label.domains": "Domain",
```
rhtyd, line 707: don't remember keys even  if they look redundant @davidjumani - it's used in section/titles and few other places

davidjumani: It is a duplicate key, L:716

rhtyd: ah okay good catch; maybe we should sort and de-dupe the list.

src/locales/en.json
```diff
@@ -29,6 +29,7 @@
 "inline": "Inline",
 "label.about": "About",
 "label.about.app": "About CloudStack",
+"label.accept": "Accept",
```
rhtyd, line 32: where is this used @davidjumani ?

src/views/storage/ResizeVolume.vue
```diff
@@ -24,7 +24,7 @@
             initialValue: selectedDiskOfferingId,
             rules: [{ required: true, message: 'Please select an option' }]}]"
           :loading="loading"
-          :placeholder="$t('label.offering.type')"
+          :placeholder="$t('label.diskoffering.type')"
```
rhtyd, line 27: See the param is `diskofferingid`, so instead of this we can simply call it `label.diskoffering`

src/locales/en.json
```diff
@@ -686,6 +687,7 @@
 "label.diskkbsread": "Disk Read (Bytes)",
 "label.diskkbswrite": "Disk Write (Bytes)",
 "label.diskoffering": "Disk Offering",
+"label.diskoffering.type": "Disk Offering Type",
```
rhtyd, line 690: Not necessary, see comment below around the usage.

src/locales/en.json
```diff
@@ -1019,6 +1020,7 @@
 "label.iso.id": "ISO ID",
 "label.iso.name": "ISO name",
 "label.isoid": "ISO",
+"label.isolated": "Isolated",
```
rhtyd, line 1023: There is a way to avoid this, will check tomorrow.

src/views/offering/AddNetworkOffering.vue
```diff
@@ -273,7 +273,7 @@
         <a-form-item :label="$t('label.tags')">
           <a-input
             v-decorator="['tags', {}]"
-            :placeholder="this.$t('label.networktags')"/>
```
rhtyd, line 276: We can simply call it tags - note the :label already uses label.tags.

src/locales/en.json
```diff
@@ -1277,6 +1280,7 @@
 "label.network.offering.name": "Network Offering Name",
 "label.network.offerings": "Network Offerings",
 "label.network.service.providers": "Network Service Providers",
+"label.network.tags": "Network Tags",
```
rhtyd, line 1283: Not necessary.

src/locales/en.json
```diff
@@ -1068,6 +1070,7 @@
 "label.kubernetesversionid": "Kubernetes version",
 "label.kubernetesversionname": "Kubernetes version",
 "label.kvmnetworklabel": "KVM Traffic Label",
+"label.l2": "L2",
```
rhtyd, line 1073: same as above

src/locales/en.json
davidjumani: https://github.com/apache/cloudstack-primate/blob/master/src/views/offering/AddNetworkOffering.vue#L108

src/locales/en.json
```diff
@@ -1426,6 +1432,7 @@
 "label.portableipaddress": "Portable IPs",
 "label.portforwarding": "Port Forwarding",
 "label.powerstate": "Power State",
+"label.preferred": "Prefered",
```
rhtyd, line 1435: Where is this used?

src/locales/en.json
davidjumani: Done!

src/locales/en.json
```diff
@@ -1539,6 +1549,7 @@
 "label.region": "Region",
 "label.region.details": "Region details",
 "label.reinstall.vm": "Reinstall VM",
+"label.reject": "Reject",
```
rhtyd, line 1552: check old UI, for add network offering; if reject is used or something else?

src/views/storage/ResizeVolume.vue
davidjumani: Done

src/locales/en.json
davidjumani: https://github.com/apache/cloudstack-primate/blob/master/src/views/offering/AddComputeOffering.vue#L434

src/locales/en.json
davidjumani: Was reject itself, but as an inline string

src/locales/en.json
rhtyd: Sorry older master, that's why I didn't catch it with git grep

PR #426, opened 2020-06-16
compute: import vm

shwstppr:
Adds UI for importing VM
Importing VM is currently supported only on VMware.

### DIFF
--- a/src/config/section/compute.js
+++ b/src/config/section/compute.js
@@ -73,6 +73,14 @@ export default {
           component: () => import('@/views/compute/DeployVM.vue')
         },
         {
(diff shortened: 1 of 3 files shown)


src/config/section/compute.js
```diff
@@ -72,6 +72,14 @@ export default {
           listView: true,
           component: () => import('@/views/compute/DeployVM.vue')
         },
+        {
+          api: 'importUnmanagedInstance',
+          icon: 'down-square',
+          label: 'label.vm.import',
+          listView: true,
+          popup: true,
```
rhtyd, line 80: show/guard/show only for VMware

PR #430, opened 2020-06-17
add custom form-item

shwstppr:
Add a custom widget with information popup button to use it across forms.

**NOT WORKING**


rhtyd:
@shwstppr there's a way to make any new widget/component available system-wide so you won't need to import in all forms; see core bootstrap

shwstppr:
Closing for now, will look into later if needed

PR #431, opened 2020-06-18
Add bypassvlanoverlapcheck parameter when create private gateway

### LINKED ISSUE #403: [BUG] Check and fix bypassvlanoverlapcheck for VPC private gw
Check against 
https://github.com/apache/cloudstack/pull/3899

davidjumani:
Fixes #403

### DIFF
--- a/src/config/section/network.js
+++ b/src/config/section/network.js
@@ -286,13 +286,12 @@ export default {
           label: 'label.add.private.gateway',
           listView: true,
           args: (record, store) => {
-            var fields = ['physicalnetworkid', 'vlan', 'ipaddress', 'gateway', 'netmask', 'sourcenatsupported', 'aclid']
+            var fields = ['vpcid', 'physicalnetworkid', 'vlan', 'ipaddress', 'gateway', 'netmask', 'sourcenatsupported', 'aclid']
             if (store.apis.createPrivateGateway.params.filter(x => x.name === 'bypassvlanoverlapcheck').length > 0) {
               fields.push('bypassvlanoverlapcheck')
             }
             return fields
           },
-
           mapping: {
             aclid: {
               api: 'listNetworkACLLists'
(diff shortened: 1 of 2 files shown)


src/config/section/network.js
```diff
@@ -285,7 +285,8 @@ export default {
           icon: 'plus',
           label: 'label.add.private.gateway',
           listView: true,
-          args: ['physicalnetworkid', 'vlan', 'ipaddress', 'gateway', 'netmask', 'sourcenatsupported', 'aclid'],
+          // TODO : Add version check
+          args: ['physicalnetworkid', 'vlan', 'bypassvlanoverlapcheck', 'ipaddress', 'gateway', 'netmask', 'sourcenatsupported', 'aclid'],
```
rhtyd, line 289: @davidjumani no need, instead use the syntax to return args from a method (see example where args can be a `() => { return ...`). Basically you can use the example I showed you yesterday to check if param is available in the API to include in the args keys or no (i.e. check API discovery api.params if it exists, if yes return otherwise don't).

src/views/network/VpcTab.vue
```diff
@@ -130,6 +130,12 @@
                   v-decorator="['vlan', {rules: [{ required: true, message: `${this.$t('label.required')}` }]}]"
                 ></a-input>
               </a-form-item>
+              // TODO : Add version check
+              <a-form-item :label="$t('label.bypassvlanoverlapcheck')">
```
rhtyd, line 134: Add check to see if API param is available, @davidjumani see https://github.com/apache/cloudstack-primate/commit/e606fe93560bc7b7d45baaa7b9193a4d60ab8b86

davidjumani:
@blueorangutan package

davidjumani:
@rhtyd Good to go! Also fixed missing vpcid when trying to create a private gateway

PR #433, opened 2020-06-18
Fix show template settings details

### LINKED ISSUE #393: [BUG] Vmware default centos template settings not populating
For the default VMware centos template, setting details are not populating (drop-down not seen in Primate) like the legacy UI:
Screenshot from 2020-06-08 09-09-32

davidjumani:
Fixes #393


davidjumani:
@blueorangutan package

davidjumani:
@rhtyd Good to go! Also fixed when removing all settings details

PR #436, opened 2020-06-19
feature: template ovf properties

### LINKED ISSUE #374: [FEATURE] a query component for text driven questions to a user
**Is your feature request related to a problem? Please describe.**
When a user uploads an OVA template it might contain properties that require answering when deploying a VM based on that template. This is a concrete example but there might be others where the UI needs to query the user based on a text input. in the OVA deployment case we have
- a query
- a name
- a default value
- a type and
- a constraints field (called "qualifiers" internally)
I think a component might have more use than the OVA deployment. It can be used for cmdline parameters or more specific control of user-data input. Maybe for some plugins it could be used to control hardware specific configuration. I think it could even be useful for more static parts of the UI interaction.

**Describe the solution you'd like**
I would like a vue.js component that takes above input and can return the user choice or input (depending on type and constraints).

**Describe alternatives you've considered**
An alternative would be javascript methods as implemented in the old UI's instanceWizard. The amount of work would be the same but it would not be usable in multiple location but as boilerplate code.

**Additional context**
In the old UI a step 7.5 of sorts was put between step 7 and 8 of the deploy wizard in case an OVA would contain ovf properties. in the new UI this would be part of the deployment form (mistakenly called a wizard still).

shwstppr:
Fixes #374 
Screenshot from 2020-06-19 13-36-45

When a user uploads an OVA template it might contain properties that require answering when deploying a VM based on that template. This is a concrete example but there might be others where the UI needs to query the user based on a text input. in the OVA deployment case we have

-a query
-a name
-a default value
-a type and
-a constraints field (called "qualifiers" internally)

I think a component might have more use than the OVA deployment. It can be used for cmdline parameters or more specific control of user-data input. Maybe for some plugins it could be used to control hardware specific configuration. I think it could even be useful for more static parts of the UI interaction.


src/views/compute/DeployVM.vue
```diff
@@ -761,6 +825,9 @@ export default {
     }
   },
   methods: {
+    removeQuotes (value) {
```
rhtyd, line 828: JSON.stringify()  (javascript pojo)

src/views/compute/DeployVM.vue
```diff
@@ -1209,6 +1284,34 @@ export default {
         this.loading.isos = false
       })
     },
+    fetchTemplateOvfProperties (templateId) {
+      this.loading.templateOvfProperties = true
+      const params = {
+        id: templateId
+      }
+      api('listTemplateOvfProperties', params).then(json => {
```
rhtyd, line 1292: Note - upstream PR/feature may get rid of this API; and instead return the properties as part of the listTemplates API response cc @DaanHoogland

rhtyd:
@blueorangutan package

rhtyd:
cc @DaanHoogland Abhishek is done with the UI component, based on any changes in the API/response we may need to refactor this again (should be quicker now that we've the groundwork complete)

rhtyd:
@blueorangutan package

rhtyd:
Please merge latest master and fix any conflicts? Thanks.
@davidjumani please take over and discuss with @DaanHoogland for any changes in the API/response done in backend.

rhtyd:
With @davidjumani

rhtyd:
With @Pearl1594

rhtyd:
Update - discussed with @DaanHoogland I'll review the backend PR and I'll make further changes in this PR

rhtyd:
Tested with backend, will do a final test by tomorrow and merge.

PR #441, opened 2020-06-22
autogenview: fix fill edit/update forms

### LINKED ISSUE #417: [BUG] Host related issues
- Go to Infra-> host -> edit host -> save - it fails.

shwstppr:
Fix update host form issue (update forms in general)

Fixes #417

### DIFF
--- a/src/views/AutogenView.vue
+++ b/src/views/AutogenView.vue
@@ -740,7 +740,8 @@ export default {
       this.currentAction.paramFields.map(field => {
         let fieldValue = null
         let fieldName = null
-        if (field.type === 'list' || field.name === 'account' || (this.currentAction.mapping && field.name in this.currentAction.mapping)) {
+        if (field.type === 'list' || field.name === 'account' ||
+          (this.currentAction.mapping && field.name in this.currentAction.mapping && !(field.name in this.resource))) {
           fieldName = field.name.replace('ids', 'name').replace('id', 'name')
         } else {
           fieldName = field.name
--- a/src/views/AutogenView.vue
+++ b/src/views/AutogenView.vue
@@ -740,8 +740,7 @@ export default {
       this.currentAction.paramFields.map(field => {
         let fieldValue = null
         let fieldName = null
-        if (field.type === 'list' || field.name === 'account' ||
-          (this.currentAction.mapping && field.name in this.currentAction.mapping && !(field.name in this.resource))) {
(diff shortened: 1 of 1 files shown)


rhtyd:
Can you explain what is the cause and fix @shwstppr ?

shwstppr:
@rhtyd code setting name of entity instead of ID in a-select. In updateHost form it sets oscategoryname value in the form element instead of oscategoryid.
Maybe to further enhance prefilling edit forms we can get the id from select list if the id param is noy available. But that might be little tricky.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

shwstppr:
@rhtyd AddHost form seems to be working fine. The only difference from legacy UI is that the form closes only after host addition result is received. In legacy UI, the form used to get closed instantly after clicking OK.
Also, does the current change look okay to you? We can open PR for review if it is.

src/views/AutogenView.vue
```diff
@@ -740,7 +740,8 @@ export default {
       this.currentAction.paramFields.map(field => {
         let fieldValue = null
         let fieldName = null
-        if (field.type === 'list' || field.name === 'account' || (this.currentAction.mapping && field.name in this.currentAction.mapping)) {
+        if (field.type === 'list' || field.name === 'account' ||
+          (this.currentAction.mapping && field.name in this.currentAction.mapping && !(field.name in this.resource))) {
```
rhtyd, line 744: actually a field name could have a custom mapping and be in a resource (there are a few example), so this may break such cases; can you fix it by rather adding a mapping to the edit host action @shwstppr but retain the `this.currentAction.mapping && field.name in this.currentAction.mapping` check

shwstppr: @rhtyd we already have a mapping for `oscategoryid` in `hosts.js` which is why we are seeing this problem.
https://github.com/apache/cloudstack-primate/blob/master/src/config/section/infra/hosts.js#L63-L67
While prefilling current values of params in edit forms, code is replacing `oscategoryid` to `oscategoryname` and using the value for `oscategoryname` from the resource.
IMO, it should not do this replacement just because we have a mapping for a param but since I'm not sure if removing it completely will break some case I just added an additional condition to stop this replacement when the original param (`oscategoryid` in `listHost` response) is available in the resource object.

PR #446, opened 2020-06-23
Allow creating Compute/Disk offering as Domain admin

ravening:
Currently only "ROOT" admin can create compute and
disk offering. If we login as domain admin then we
cant create a compute and disk offering as domain
id's is not passed to the api. If we login as
domain admin we need to set "isPublic" to false
so that domain id list will be displayed

Before the changes:

This is the error we get for compute/disk offering creation

Screenshot 2020-06-23 at 10 53 45

After the changes:

The UI for copute/disk offering creation shows the domain id list

Screenshot 2020-06-23 at 10 59 44

Also we no longer need to use "isrecursive=true" since it wont display the compute/disk offering belonging to parent domain in the child domain

Let say we have parent domain "test1" and its child domain "test-11" then all the offerings created by "test1" will not be visible to its child domain "test-11" if isrecursive is true

Screenshot 2020-06-23 at 11 16 19

Setting isrecursive to false will display all the parent domain offerings

Screenshot 2020-06-23 at 11 16 44

### DIFF
--- a/src/views/offering/AddComputeOffering.vue
+++ b/src/views/offering/AddComputeOffering.vue
@@ -595,6 +595,7 @@ export default {
       this.isSystem = true
     }
     this.fetchData()
+    this.isPublic = this.isAdmin()
   },
   methods: {
     fetchData () {
--- a/src/views/offering/AddDiskOffering.vue
+++ b/src/views/offering/AddDiskOffering.vue
@@ -348,6 +348,7 @@ export default {
   },
   mounted () {
     this.fetchData()
+    this.isPublic = this.isAdmin()
   },
   methods: {
     fetchData () {


rhtyd:
@shwstppr can you review and test this?
@blueorangutan package

src/config/section/offering.js
```diff
@@ -27,7 +27,6 @@ export default {
       docHelp: 'adminguide/service_offerings.html#compute-and-disk-service-offerings',
       icon: 'cloud',
       permission: ['listServiceOfferings', 'listDomains'],
-      params: { isrecursive: 'true' },
```
shwstppr, line 30: this is not correct. `isrecursive` must be passed to list offerings for domain and sub domain. `isrecursive=true` is inline with legacy UI
Screenshot from 2020-06-23 16-14-39

src/config/section/offering.js
```diff
@@ -109,7 +108,6 @@ export default {
       icon: 'hdd',
       docHelp: 'adminguide/service_offerings.html#compute-and-disk-service-offerings',
       permission: ['listDiskOfferings', 'listDomains'],
-      params: { isrecursive: 'true' },
```
shwstppr, line 112: same as above

src/views/offering/AddComputeOffering.vue
```diff
@@ -606,6 +606,9 @@ export default {
       }
     },
     isAdmin () {
+      if (!['Admin'].includes(this.$store.getters.userInfo.roletype)) {
+        this.isPublic = false
+      }
```
shwstppr, line 611: `isAdmin` is a getter function and should not be setting variable values.
Also, this change does not solve root issue, ie, domain selection element is not visible to domain admin in create Compute Offering offering and isPublic is passed true in create API.
Screenshot from 2020-06-23 16-19-04
Screenshot from 2020-06-23 16-19-30

src/views/offering/AddDiskOffering.vue
```diff
@@ -358,6 +358,9 @@ export default {
       }
     },
     isAdmin () {
+      if (!['Admin'].includes(this.$store.getters.userInfo.roletype)) {
+        this.isPublic = false
+      }
```
shwstppr, line 363: Same as above.
Instead setting isPublic to false here. `handleSubmit` should be checked for the value and visibility of `a-select` element for domain must be fixed for domain admin

shwstppr: @ravening correct way, in my opinion, would be adding following line in `mounted()` method of `AddComputeOffering.vue` and `AddDiskOffering.vue`,
`
this.isPublic = this.isAdmin()`

src/config/section/offering.js
ravening: `isrecursive=true` has a bug with disk/compute offerings

It wont display the parent offerings in the child domain. This bug has to fixed in cloudstack codebase as well

Lets say I have ROOT -> child -> grandchild domains and if I created offering for "child" domain then I cant view that in "grandchild" domain using isrecursive=true

ravening: @shwstppr 

Lets have I have three domains like "ROOT" -> "ROOT/test1" -> "ROOT/test1/test11"

With `isrecursive=true` i dont see all the disk offering of domain "ROOT/test1" from "ROOT/test1/test11"

Screenshot 2020-06-23 at 16 14 09

with `isrecursive=false` then I do see all offerings of "ROOT/test1" from "ROOT/test1/test11"

Screenshot 2020-06-23 at 16 14 27


The cloudmonkey also shows same

```
(test11) 🐵 > list diskofferings listall=true filter=name,domain isrecursive=true
{
  "count": 5,
  "diskoffering": [
    {
      "name": "Small"
    },
    {
      "name": "Medium"
    },
    {
      "name": "Large"
    },
    {
      "name": "Custom"
    },
    {
      "domain": "/test1/test-11/",
      "name": "test-11only"
    }
  ]
}
```

With isrecursive=false

```
(test11) 🐵 > list diskofferings listall=true filter=name,domain
{
  "count": 8,
  "diskoffering": [
    {
      "name": "Small"
    },
    {
      "name": "Medium"
    },
    {
      "name": "Large"
    },
    {
      "name": "Custom"
    },
    {
      "domain": "/test1/",
      "name": "test1only"
    },
    {
      "domain": "/test1/test-11/",
      "name": "test-11only"
    },
    {
      "domain": "/test1/",
      "name": "test1-and-11"
    },
    {
      "domain": "/",
      "name": "root-test1-test11"
    }
  ]
}
```

ravening: So either we need to retain isrecursive=true and fix the backend code or detain pass isrecursive

src/views/offering/AddDiskOffering.vue
ravening: Yes this is fine. I will make changes for this part

src/config/section/offering.js
shwstppr: With current state I don't think we can show parent domain offerings in this view. It will allow showing editing UI for those offerings although user doesn't have access to edit them.
As far as I remember, legacy UI allow showing only current and subdomain offerings in this view and while all accessible offering in deplovm view where isrecursive as true. In any of these better option would be to add an issue in CloudStack itself. @ravening

ravening: @shwstppr In the legacy Ui also its a bug and the sub child domain cant see the offerings of the parent domain which is the child of "ROOT" domain.
Anyway I will retain the code here and create an issue in cloudstack

rhtyd:
@blueorangutan package

rhtyd:
@shwstppr can you re-review the PR and also check against old UI and if that's wrong what ideal parameters must be sent as part of API request 
@blueorangutan package

shwstppr:
@rhtyd already LGTM'd the latest changes. Now it only fixes the create form

ravening:
@rhtyd @shwstppr its still an issue in the old UI as well. The java code looks good but we see a different behavior in UI. If thats the case we need to fix it in UI right? Raising an issue with cloudstack codebase or here doesnt matter

PR #454, opened 2020-06-24
autogenview: store API parameters as router query for history

### LINKED ISSUE #451: Maintain page context in the URL
The page context for page, pagesize, keyword (maybe filter related things later on...) are not stored on the route/path/params and back/forward missing the context...

rhtyd:
This would store pagination, filter and search context in route query
which would allow for browser history navigation in list views.

Fixes #451

Signed-off-by: Rohit Yadav <rohit.yadav@shapeblue.com>


rhtyd:
Please test and review @davidjumani

rhtyd:
@blueorangutan package

rhtyd:
Converted to draft, as this may not be complete yet
With @davidjumani

rhtyd:
@davidjumani no update/progress on this?

davidjumani:
Have tested for VMs, Images and Networks. Testing the rest in progress

src/views/AutogenView.vue
```diff
@@ -381,10 +382,19 @@ export default {
   watch: {
     '$route' (to, from) {
       if (to.fullPath !== from.fullPath && !to.fullPath.includes('action/')) {
-        this.searchQuery = ''
-        this.page = 1
+        if ('q' in to.query) {
```
davidjumani, line 385: should pick up `query.keyword` instead ?

rhtyd: I think not all views/APIs pass keyword as the query parameter (from search); for example global setting we use name

davidjumani: Agreed but it creates two parameters keywork / name and q, with q having the higher priority, so changing only q and not the other can lead to inconsistencies. It by default picks up keywork / name when q is not present anyway

davidjumani:
TODO once done, add the functionality to redirect to running / stopped vms via dasboard

rhtyd:
@blueorangutan package

rhtyd:
Hold-on merging this, we may need more tests.

rhtyd:
Ready for review and testing @davidjumani 
@blueorangutan package

rhtyd:
Please review/test @davidjumani @Pearl1594 thanks

rhtyd:
Did another round of test, please continue testing master instead @Pearl1594 @davidjumani

PR #459, opened 2020-06-25
Display tab showing the current resource count for domain

ravening:
Currenlty "Domain Admins" cannot see how much resources
they are consuming for their domain.
So display a "Resource Count" tab which displays the
current usage and the total allocated resoures

Screenshot 2020-06-25 at 17 11 08

Also fix the display error for "parentdomainname" label

Screenshot 2020-06-25 at 15 29 20


rhtyd:
@blueorangutan package

rhtyd:
@ravening here some feedback and ideas:
- Rename the tab to Resource or something like Usage?
- Could you make it reusable, as it's the same concept for account and projects (similar to the limits tab)
- If we need to show the current resource count/usage in a different tab, remove it from the details tab (do it for accounts and projects as well)
- See Infra -> zone -> resources tab which shows capacities/resource count for zones; you may even show progress bar like widgets (we know current and total usage, to show a percentage for example)

ravening:
> @ravening here are some feedback and ideas:
> 
> * Rename the tab to Resource or something like Usage?
> * Could you make it reusable, as it's the same concept for account and projects (similar to the limits tab)
> * If we need to show the current resource count/usage in a different tab, remove it from the details tab (do it for accounts and projects as well)
> * See Infra -> zone -> resources tab which shows capacities/resource count for zones; you may even show progress bar like widgets (we know current and total usage, to show a percentage for example)

@rhtyd 
Im not sure how to display the progress bar properly as most of the resources will be "-1" for root domain and other domains. In that case what value should I use for denominator?

In the details tab they just iterate over the values in the array. So it might be difficult to remove the details from there

rhtyd:
@ravening we can skip those or show a progress bar with no progress (0%) as the limit is unlimited?

ravening:
@rhtyd I made the proper changes showing progress bar as well for Domains/accounts/projects
Please review again

Screenshot 2020-07-02 at 12 21 37

davidjumani:
@blueorangutan package

davidjumani:
@ravening This looks good, can you treat -1 as unlimited and display it accordingly?

ravening:
> @ravening This looks good, can you treat -1 as unlimited and display it accordingly?

@davidjumani I didnt see "-1" in the output of listDomains or listAccounts or listProjects. It returns "Unlimited" instead of -1

davidjumani:
@ravening Perhaps it's an older version, could you handle this too ?
Screenshot from 2020-07-03 19-09-21

rhtyd:
LGTM

PR #460, opened 2020-06-26
Fix 403 on login

### LINKED ISSUE #456: [BUG] After session expire, upon new login user is given 403
**Describe the bug**
After session expire and following log, user is being displayed with 403 instead of the resource he had just authenticated to. 

**To Reproduce**
1. Let the session expire while on a page
2. Click anywhere on the screen
3. You're being redirected to login, login to CS
4. Instead of the page you've clicked before login you've been displayed with 403

**Expected behavior**
User should be redirected to the page where he was last

davidjumani:
Fixes #456 
However, this might lead to longer wait time after login if there is a redirect.
Could also check and remove Dashboard from the redirect

### DIFF
--- a/src/permission.js
+++ b/src/permission.js
@@ -51,7 +51,7 @@ router.beforeEach((to, from, next) => {
           message.loading('Loading...', 1.5)
         }
         store
-          .dispatch('GetInfo')
+          .dispatch('GetInfo', from.query.redirect && from.query.redirect !== '/dashboard')
           .then(apis => {
             store.dispatch('GenerateRoutes', { apis }).then(() => {
               router.addRoutes(store.getters.addRouters)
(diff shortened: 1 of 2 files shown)


davidjumani:
@blueorangutan package

rhtyd:
@borisstoyanov can you review this - while it now goes back to the page/view you were on (before logout or session expiry) the login would appear slow/delayed. Trade off is -> whether to show dashboard directly (i.e. user will feel login is fast enough) or go to the page before logout/session-timeout but login would appear slow

rhtyd:
Thought it over last few days, I think the feature to allow users to be redirected to a url outweighs the "feel" that UI is loading faster.
Reverted my change from the UX change in https://github.com/apache/cloudstack-primate/commit/960ba1bd766c11886e6023ecd5ba5fd278ff6a5e

PR #461, opened 2020-06-26
Fixing change order available to read only admin

### LINKED ISSUE #457: [BUG] Change order of Service offering is available to Read-only admin
**Describe the bug**
With the latest default role introduced - read only admin he's able to see the buttons for changing the order of * offerings
When click on some of the buttons he's immediately logged out. 

As per the autodisovery nature of this UI I believe users shouldn't be able to see such buttons, not sure if there's a solution on a framework level but it'll be good to look into such instead of looking into each custom role and logging each button that shouldn't be available to user. 

readonly

davidjumani:
Fixes #457

### DIFF
--- a/src/components/view/ListView.vue
+++ b/src/components/view/ListView.vue
@@ -19,7 +19,7 @@
   <a-table
     size="small"
     :loading="loading"
-    :columns="columns"
+    :columns="fetchColumns()"
     :dataSource="items"
     :rowKey="record => record.id || record.name"
     :pagination="false"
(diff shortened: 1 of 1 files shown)


shwstppr:
@blueorganutan package

shwstppr:
@blueorangutan package

davidjumani:
@blueorangutan package

src/components/view/ListView.vue
```diff
@@ -275,6 +275,12 @@ export default {
     }
   },
   methods: {
+    fetchColumns () {
```
rhtyd, line 278: Usually we would want a computed property, but this works

PR #469, opened 2020-06-29
Diabling actions for readonly admin

### LINKED ISSUE #467: [BUG] Read-only admin issues
As a read-only admin the following haven't been fixed yet:
- Readonly admin shouldn't be allowed to change order of role/rule permissions
- Don't allow changing public IP address (add/delete/change) for a physical network of zone
- Discuss - Hide console for a VM if the user/account is not allowed to deploy/stop/start VM?

davidjumani:
Fixes #467

### DIFF
--- a/src/components/widgets/Console.vue
+++ b/src/components/widgets/Console.vue
@@ -17,7 +17,7 @@
 
 <template>
   <a
-    v-if="['vm', 'systemvm', 'router', 'ilbvm'].includes($route.meta.name)"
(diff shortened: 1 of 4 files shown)


rhtyd:
@davidjumani can you add a getters check on console widget against the API `updateVirtualMachine` - the idea that if an account is not allowed to update a VM, it shouldn't be allowed to access the console

davidjumani:
@rhtyd Would that apply only to user vms or systemvms too ?

davidjumani:
@blueorangutan package

rhtyd:
@blueorangutan package

davidjumani:
@rhtyd Good to go!

Issue #473, opened 2020-06-30
[BUG] Templates screen should not show 'order' column to non-root admins


PaulAngus:
**Describe the bug**

Someone of type 'user' should not be able to see the Order column - it's not a user-supported function.

**To Reproduce**

1. Log in as a normal user
2. Go to 'Images -> templates'

Order column is visible - this should only be visible to root admins

**Expected behavior**

The Order column should not be visible to normal users.

**Screenshots**
![image](https://user-images.githubusercontent.com/4810220/86139071-348ba480-bae7-11ea-9f52-72de04b243d9.png)

**Desktop (please complete the following information):**
 - OS: [e.g. Windows, Mac, iOS, Android with device/screen details if applicable]

**Additional context**

PR #474, opened 2020-06-30
Hide order column for user account

Pearl1594:
Disables the order column in Templates List View for normal user - Addresses issue: https://github.com/apache/cloudstack-primate/issues/473

### DIFF
--- a/src/views/AutogenView.vue
+++ b/src/views/AutogenView.vue
@@ -485,10 +485,6 @@ export default {
         }
       }
 
-      if (!['Admin'].includes(this.$store.getters.userInfo.roletype) && this.apiName === 'listTemplates' && this.selectedFilter !== 'self') {
-        this.columnKeys = this.columnKeys.filter(key => { return key !== 'order' })
-      }
-
       if (this.selectedFilter && this.filters.length > 0) {
         if (this.$route.path.startsWith('/template')) {
           params.templatefilter = this.selectedFilter
--- a/src/views/AutogenView.vue
+++ b/src/views/AutogenView.vue
@@ -485,6 +485,10 @@ export default {
         }
       }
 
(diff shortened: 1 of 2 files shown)


Pearl1594:
@blueorangutan package

Pearl1594:
@blueorangutan package

rhtyd:
@blueorangutan package

src/views/AutogenView.vue
```diff
@@ -485,6 +485,10 @@ export default {
         }
       }
 
+      if (!['Admin'].includes(this.$store.getters.userInfo.roletype) && this.apiName === 'listTemplates' && this.selectedFilter !== 'self') {
```
rhtyd, line 488: @Pearl1594 do we need to check for the selected filter; shouldn't we not display the order column irrespective of that? Maybe check against old UI?

Pearl1594: @rhtyd In the old UI - the order column is shown for all users , but the backend API prevents it from ordering them with the following exception: 
```
Domain Admin and regular users can modify only their own Public templates
```
Hence, to adhere to this - I've exposed the order column only for those that it owns

rhtyd: @Pearl1594 so if an account owns templates, could they update and change order for their own (public) templates? Instead, we can just enforce that the order columns are shown only to the root admin.

Pearl1594: The API allows users and dom admins to change the order of their own public templates. However, I can restrict this only to Root admin

davidjumani:
@blueorangutan package

rhtyd:
@blueorangutan package

davidjumani:
@blueorangutan package

Pearl1594:
@blueorangutan package

Pearl1594:
Yup, after the api is executed, the order gets reset

PR #493, opened 2020-07-05
Internationalize state output

nvazquez:
CloudStack API methods return resources states in English, switching Primate to a different language should also translate these states.

Example: VM states: Running, Stopped, Migrating,...

image
image

### DIFF
--- a/src/components/widgets/Status.vue
+++ b/src/components/widgets/Status.vue
@@ -38,7 +38,37 @@ export default {
   methods: {
     getText () {
       if (this.displayText && this.text) {
-        return this.text.charAt(0).toUpperCase() + this.text.slice(1)
+        var state = this.text
+        switch (state) {
+          case 'Running':
+            state = this.$t('state.running')
+            break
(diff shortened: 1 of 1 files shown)


src/components/widgets/Status.vue
```diff
@@ -38,7 +38,37 @@ export default {
...
+            state = this.$t('state.migrating')
+            break
+          case 'Expunging':
+            state = this.$t('state.expunging')
+            break
+          case 'Error':
+            state = this.$t('state.error')
+            break
+        }
+        return state.charAt(0).toUpperCase() + state.slice(1)
```
rhtyd, line 71: In non-English languages, this might cause an issue? (many languages don't have notion of upper vs lower case)

nvazquez:
@blueorangutan package

src/components/widgets/Status.vue
nvazquez: Have tested every language manually on Chrome and Firefox and found no issues.

PR #514, opened 2020-07-08
Add missing internationalization on resources tags and annotations

nvazquez:
Missing internationalization on tags and annotations


nvazquez:
@blueorangutan package

PR #515, opened 2020-07-08
Add missing internationalization on instance groups view

nvazquez:
Add missing internationalization on instance groups view


nvazquez:
@blueorangutan package

PR #524, opened 2020-07-09
offerings: Removing tags for offerings

davidjumani:
Removing tags from offerings as they do not support them


davidjumani:
@blueorangutan package

rhtyd:
@davidjumani please check against old UI, while we may not support this in UI - the API supports the tags parameter

ravening:
@davidjumani sometimes we do use tags for offerings to manage vm deployments and vm migrations depending on offering and host tags. So removing may not be a good idea

rhtyd:
@ravening are you okay with latest changes?
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
We need one additional review, pinging @utchoang @Pearl1594 thanks.

davidjumani:
It doesn't show / add tags in the project view, but can in the default view (only for vpc offerings). Adding resourcetype for vpc offerings. Will need another PR to handle it in case of projects

davidjumani:
@blueorangutan package

PR #525, opened 2020-07-09
Display URL for the template to admins

ravening:
In legacy UI, we display the url of the template to
admins. Do the same thing here also

Below is the view for admin

Screenshot 2020-07-09 at 11 25 07

And below is the view for domain admins/users

Screenshot 2020-07-09 at 11 24 04

### DIFF
--- a/src/config/section/image.js
+++ b/src/config/section/image.js
@@ -42,7 +42,7 @@ export default {
         }
         return fields
       },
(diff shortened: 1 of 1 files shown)


rhtyd:
The listTemplates response does not return URL, but we can still accept the change as if it does we can display it
@blueorangutan package

rhtyd:
Confirmed URL not seen is API does not return the key. Can merge as probably some CloudStack forks maybe returning the URL as part of list API response.

Issue #520, opened 2020-07-09
[BUG] DeployVM Wizard does not properly search for templates


davidjumani:
**Describe the bug**
In the deploy VM wizard, the search only searches the current selection of templates and does not query the backend

**To Reproduce**

1. Log in as a User or an Admin
2. Go to Create new VM
3. Search for a template

**Expected behavior**
An api call with the keyword should be called

rhtyd:
@utchoang can you have a look at this, we want:
- The search to actually search the template/iso API
- Render the results on their own specific tabs (i.e. filtered tabs of features, public, mine... etc)
- On clicking the radio button we see an oval animation (minor issue, probably flex or css related)

ravening:
@rhtyd @davidjumani 

I tried adding

```
@handle-search-filter="($event) => handleSearchFilter('templates', $event)"
```
under `<a-card` for the template/iso section but still couldn't search using the api. Mybe im missing something

rhtyd:
@davidjumani please address this issue on priority as it's blocking general usability, in addition:
- As a user I should be able to search and select template based on the 4 filters (tabs)
- Unless I override the root disk option, it should not force or set the rootdisk value (it should also change if I change the template, or reset if I select an ISO)
- The rootdisk option shouldn't be passed via API if the override is/was not selected, also the value must be a whole number (than floating point). By default in VMware where I tested this, it is breaking VM deployment using Primate.

I see there is a PR but may not address all the issues in which case you may take that may as reference or cherry-pick commit and send a new PR. Thanks.

PR #527, opened 2020-07-09
Enable searching for templates/iso while vm deployment

ravening:
Fix for #520 

Curertly we cant search for templates using API while
deploying a VM. It searches for local cache and not from
API server. Allow searching for templates from backend
using filters

Need the same functionality for iso as well

Listing the templates after selecting the zone

Screenshot 2020-07-09 at 18 46 40

Screenshot 2020-07-09 at 18 46 47
Screenshot 2020-07-09 at 18 46 54
Screenshot 2020-07-09 at 18 46 59

After searching for a particular template

Screenshot 2020-07-09 at 18 47 12
Screenshot 2020-07-09 at 18 48 14

### DIFF
--- a/src/views/compute/wizard/TemplateIsoSelection.vue
+++ b/src/views/compute/wizard/TemplateIsoSelection.vue
@@ -54,6 +54,8 @@
 import { getNormalizedOsName } from '@/utils/icons'
 import TemplateIsoRadioGroup from '@views/compute/wizard/TemplateIsoRadioGroup'
 import store from '@/store'
+import { api } from '@/api'
+import _ from 'lodash'
 
 export default {
   name: 'TemplateIsoSelection',
@@ -143,10 +145,10 @@ export default {
       }
       this.filteredItems.forEach((os) => {
         os.osName = getNormalizedOsName(os.ostypename)
-        if (os.isPublic && os.isfeatured) {
+        if (os.ispublic && !os.isfeatured) {
           mappedItems.community.push(os)
           itemCount.community = itemCount.community + 1
-        } else if (os.isfeatured) {
+        } else if (os.ispublic && os.isfeatured) {
           mappedItems.featured.push(os)
           itemCount.featured = itemCount.featured + 1
         } else {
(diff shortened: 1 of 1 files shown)


utchoang:
@ravening @rhtyd I think we should fix it as follows:
- Regarding fetch data or search template/isos, you should call API by tab instead of fetching all as currently.
- When clicking on a tab, call API to fetch data according to that tab instead of the original filter.

ravening:
> @ravening @rhtyd I think we should fix it as follows:
> 
> * Regarding fetch data or search template/isos, you should call API by tab instead of fetching all as currently.
> * When clicking on a tab, call API to fetch data according to that tab instead of the original filter.

@utchoang I think its better to fetch once and load all data. If we do api call for each tab then the user experience may not good if fetching the templates for each tab takes time.

If we fetch all templates initially, it might take some time but there is no lag while changing the tabs.
This is the default behavior in the legacy ui also

rhtyd:
@blueorangutan package

ravening:
@rhtyd @davidjumani please review the changes

rhtyd:
Can you help review @utchoang @davidjumani thanks
@blueorangutan package

rhtyd:
@davidjumani can you review/test this as well? Thnx

src/views/compute/wizard/TemplateIsoSelection.vue
```diff
@@ -232,6 +235,57 @@ export default {
...
+      }
+      // Fetch templates/isos for all filters
+      templateFilter.forEach(filter => {
+        params.templatefilter = filter
+        promises.push(this.fetchTemplatesIsos(apiCommand, keyword, params))
+      })
+
+      Promise.all(promises).then(response => {
+        response.forEach((resItem) => {
+          const concatTemplates = _.concat(templates, _.get(resItem, apiCommandResponse, []))
```
rhtyd, line 271: @ravening with latest 4.14 that is what we use reference env, templates returned are unque when the API is called with a `showunique=true` param. This will not be necessary then, also use of lodash.

ravening: if `_.concat(templates, _.get(resItem, apiCommandResponse, []))` is not needed then how can I extract response from `resItem`?

rhtyd:
@davidjumani can you take over and help fix this, if necessary create a new PR based on this. This is a blocker if user is not able to find/select their own templates using the tabs.

PR #534, opened 2020-07-13
Bug fixes

davidjumani:
Fixes : 
- VR router offering link leads to an empty page
- In the event listview, clicking the username redirects to the event not the user
- While changing a systemvm service offering, need to fetch system offerings by passing issystem=true, virtualmachineid and systemvmtype to the listServiceOfferings api
- When editing a role, resource admin is not in the dropdown
- In DeployVM, disk offering search doesn't show 'No records' if none present but keeps the old rows
- In add ACL Rule, the Rule# field should be a number, not text
- No pagination for security groups ingress / egress rules

### DIFF
--- a/src/components/view/InfoCard.vue
+++ b/src/components/view/InfoCard.vue
@@ -374,7 +374,8 @@
           <div class="resource-detail-item__label">{{ $t('label.serviceofferingname') }}</div>
           <div class="resource-detail-item__details">
             <a-icon type="cloud" />
-            <router-link v-if="$router.resolve('/computeoffering/' + resource.serviceofferingid).route.name !== '404'" :to="{ path: '/computeoffering/' + resource.serviceofferingid }">{{ resource.serviceofferingname || resource.serviceofferingid }} </router-link>
+            <router-link v-if="$route.meta.name === 'router'" :to="{ path: '/computeoffering/' + resource.serviceofferingid, query: { issystem: true } }">{{ resource.serviceofferingname || resource.serviceofferingid }} </router-link>
+            <router-link v-else-if="$router.resolve('/computeoffering/' + resource.serviceofferingid).route.name !== '404'" :to="{ path: '/computeoffering/' + resource.serviceofferingid }">{{ resource.serviceofferingname || resource.serviceofferingid }} </router-link>
             <span v-else>{{ resource.serviceofferingname || resource.serviceofferingid }}</span>
           </div>
         </div>
--- a/src/views/network/IngressEgressRuleConfigure.vue
+++ b/src/views/network/IngressEgressRuleConfigure.vue
@@ -80,8 +80,8 @@
       style="overflow-y: auto"
       :columns="columns"
       :dataSource="rules"
-      :pagination="false"
-      :rowKey="record => record.id">
+      :pagination="{ pageSizeOptions: ['10', '20', '40', '80', '100', '500'], showSizeChanger: true}"
+      :rowKey="record => record.ruleid">
       <template slot="protocol" slot-scope="record">
         {{ record.protocol | capitalise }}
       </template>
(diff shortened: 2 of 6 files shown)


davidjumani:
@blueorangutan package

davidjumani:
@blueorangutan package

davidjumani:
@blueorangutan package

davidjumani:
@rhtyd cc @Pearl1594 Please review & merge

src/components/view/InfoCard.vue
```diff
@@ -374,7 +374,8 @@
           <div class="resource-detail-item__label">{{ $t('label.serviceofferingname') }}</div>
           <div class="resource-detail-item__details">
             <a-icon type="cloud" />
-            <router-link v-if="$router.resolve('/computeoffering/' + resource.serviceofferingid).route.name !== '404'" :to="{ path: '/computeoffering/' + resource.serviceofferingid }">{{ resource.serviceofferingname || resource.serviceofferingid }} </router-link>
+            <router-link v-if="$route.meta.name === 'router'" :to="{ path: '/computeoffering/' + resource.serviceofferingid, query: { issystem: true } }">{{ resource.serviceofferingname || resource.serviceofferingid }} </router-link>
```
rhtyd, line 377: @davidjumani should we also check for systemvm?

src/config/section/role.js
```diff
@@ -55,7 +55,7 @@ export default {
       args: ['name', 'description', 'type'],
       mapping: {
         type: {
-          options: ['Admin', 'DomainAdmin', 'User']
+          options: ['Admin', 'DomainAdmin', 'ResourceAdmin', 'User']
```
rhtyd, line 58: @davidjumani Remove ResourceAdmin, we deliberately don't want users/admins to use this role type (historic issue, related to cpbm no longer used now)

src/views/network/IngressEgressRuleConfigure.vue
```diff
@@ -79,9 +79,9 @@
       size="small"
       style="overflow-y: auto"
       :columns="columns"
-      :dataSource="rules"
+      :dataSource="rules.slice(pagesize*(page - 1), Math.min( rules.length, pagesize*page))"
```
rhtyd, line 82: @davidjumani fix parenthesis/space issue

rhtyd: Does the API not support pagination? Otherwise, I would expect to simply give all the datasource (not a slice) and instead use table's own pagination than an explicit one.

src/components/view/InfoCard.vue
davidjumani: SystemVM doesn't return the service offering

src/config/section/role.js
davidjumani: Then shall I remove it while creating a user too ?

rhtyd: @davidjumani no leave that in the createAccount form/component for legacy reasons, I think old UI shows the same list of roles as returned by listRoles.

src/components/view/InfoCard.vue
rhtyd: +1

src/views/network/IngressEgressRuleConfigure.vue
davidjumani: It calls the listSecurityGroups api and that in turn returns all the rules associated with the SG in the response, so paginating it over there

src/views/network/IngressEgressRuleConfigure.vue
```diff
@@ -79,9 +79,9 @@
       size="small"
       style="overflow-y: auto"
       :columns="columns"
-      :dataSource="rules"
+      :dataSource="rules.slice(pagesize*(page - 1), Math.min(rules.length, pagesize*page))"
       :pagination="false"
```
rhtyd, line 83: @davidjumani here if you say pagination true or remove this, would it help?

PR #539, opened 2020-07-14
Consolidated Bug fixes

davidjumani:
Fixes :
- Don't allow users in UI to delete/archive events
- Fix button name in VM deployment form in its network section to Create new network
- Refresh after template / iso upload
- Making external-id mandatory in ImportBackupOffering
- Fixing visibility of assignVirtualMachineToBackupOffering
- Removing link on traffic label
- Defensive check in TrafficTypesTab
- Ensuring we get the user info so that store.getters.user is never empty when the page is freshly loaded
- Changing from report bug to report issue
- Ordering projects in menu
- Changing router get health check results
- Show configureHAForHost based on hypervisor
- Fix scale and migrate router
- Fix scale and migrate systemvm
- Fix show actionbutton for assignVirtualMachineToBackupOffering
- Fix show actionbutton for stopKubernetesCluster
- Fix show actions for volumes
- Fix show actions for snapshots
- Fix show actions for vm snapshots
- Fix show actions for backups
- Adding loading for tags and annotations
- Enter to submit advanced search
- Fixing show Project instead of account when passed as projectaccount passed in account field
- Show project name instead of displaytext
- Fixing template and iso actions
- Fixing tags with projectid
- Fix security groups ingress/egress rules view
- Removing redundant allocationstate from zones
- Adding managedstate to clusters
- Adding capacity tab to clusters and pods
- Adding routerlink to events in dashboard
- Set autofocus to username in login


davidjumani:
@blueorangutan package

src/components/header/ProjectMenu.vue
```diff
@@ -39,7 +39,7 @@
       </a-tooltip>
 
       <a-select-option v-for="(project, index) in projects" :key="index">
-        {{ project.displaytext || project.name }}
+        {{ project.name }}
```
rhtyd, line 42: @davidjumani check against old UI

davidjumani: Fixed!

davidjumani:
@blueorangutan package

src/config/section/event.js
```diff
@@ -44,7 +44,9 @@ export default {
         ids: {
           value: (record) => { return record.id }
         }
-      }
+      },
+      show: (record, store) => { return !['User'].includes(store.userInfo.roletype) },
+      groupShow: (record, store) => { return !['User'].includes(store.userInfo.roletype) }
```
rhtyd, line 49: @davidjumani why not use the show as the implementation is same - the condition for showing the button can remain same? (or does it conflict with other views like stop/start vms etc.)?

rhtyd:
Merging this to unblock testing @davidjumani pl start another PR if new issues or regressions are seen

rhtyd:
on refresh/merge conflicts seen - @davidjumani can you fix conflicts against latest master?

src/config/section/event.js
davidjumani: Yup. Causes issues and shows stop / start based on the first vm's state

davidjumani:
@rhtyd Done!
@blueorangutan package

PR #542, opened 2020-07-15
Fixing rootdisk size override

### LINKED ISSUE #520: [BUG] DeployVM Wizard does not properly search for templates

davidjumani:
Fixes #520:
- The rootdisk option shouldn't be passed via API if the override is/was not selected, also the value must be a whole number (than floating point). By default in VMware where I tested this, it is breaking VM deployment using Primate.

### DIFF
--- a/src/views/compute/DeployVM.vue
+++ b/src/views/compute/DeployVM.vue
@@ -960,7 +960,8 @@ export default {
         })
         const templates = this.options.templates.filter(x => x.id === value)
         if (templates.length > 0) {
-          this.dataPreFill.minrootdisksize = templates[0].size / (1024 * 1024 * 1024) || 0 // bytes to GB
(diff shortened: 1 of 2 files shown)


davidjumani:
@blueorangutan package

davidjumani:
Tested and working

davidjumani:
@blueorangutan package

davidjumani:
@blueorangutan package

davidjumani:
@blueorangutan package

davidjumani:
@rhtyd good to go!

PR #553, opened 2020-07-22
Dont display "isolated network" section for SG enabled zone

ravening:
In the "add network" form, dont display the form for
isolated network if its an advanced zone with security
group enabled.

Steps:

Create an advanced zone with security groups enabled.
Login to ui, navigate to networks section and click on "Add network"

This should show only L2 and guest network option

Screenshot 2020-07-22 at 13 27 45

### DIFF
--- a/src/views/network/CreateNetwork.vue
+++ b/src/views/network/CreateNetwork.vue
@@ -99,7 +99,7 @@ export default {
           return true
         }
       }
-      return true
+      return false
     },
     handleRefresh () {
       this.fetchData()
--- a/src/views/network/CreateL2NetworkForm.vue
+++ b/src/views/network/CreateL2NetworkForm.vue
@@ -303,7 +303,7 @@ export default {
       api('listZones', params).then(json => {
         for (const i in json.listzonesresponse.zone) {
           const zone = json.listzonesresponse.zone[i]
-          if (zone.networktype === 'Advanced' && zone.securitygroupsenabled !== true) {
+          if (zone.networktype === 'Advanced') {
             this.zones.push(zone)
           }
         }


rhtyd:
@davidjumani @utchoang can you help review this? Thnx
@blueorangutan package

src/views/network/CreateNetwork.vue
```diff
@@ -95,8 +95,9 @@ export default {
     isAdvancedZoneWithoutSGAvailable () {
       for (const i in this.actionZones) {
         const zone = this.actionZones[i]
-        if (zone.networktype === 'Advanced' && zone.securitygroupsenabled !== true) {
-          return true
+        // Dont display isolated networks for advanced with sg enabled
+        if (zone.networktype === 'Advanced' && zone.securitygroupsenabled === true) {
+          return false
```
davidjumani, line 100: Would this break if there are multiple zones and one with SG Enabled ?

src/views/network/CreateNetwork.vue
```diff
@@ -95,8 +95,9 @@ export default {
...
         }
       }
       return true
```
davidjumani, line 103: return false over here instead ?

src/views/network/CreateNetwork.vue
ravening: Fixed it to match the zone id.

src/views/network/CreateNetwork.vue
```diff
@@ -95,8 +95,11 @@ export default {
     isAdvancedZoneWithoutSGAvailable () {
       for (const i in this.actionZones) {
         const zone = this.actionZones[i]
-        if (zone.networktype === 'Advanced' && zone.securitygroupsenabled !== true) {
-          return true
+        // Dont display isolated networks for advanced with sg enabled
+        if (this.resource.zoneid === zone.id &&
+          zone.networktype === 'Advanced' &&
+          zone.securitygroupsenabled === true) {
+          return false
```
davidjumani, line 102: Guessing this might still break since this.resource is passed from autogenview which is generally the first item in the list. Can leave it as before and just change the final return to false

ravening: @davidjumani  Lets say I change the code to
```
for (const i in this.actionZones) {
        const zone = this.actionZones[i]
        if (zone.networktype === 'Advanced' && zone.securitygroupsenabled !== true) {
          return true
        }
      }
      return false
```


If we have two zones, one without sg and one with sg, it will return for the first check inside the for loop right even for the zone with sg enabled?

davidjumani: Yes, so in the form the user will select the zone in which to create the network, and probably need to add a check to exclude zones without SG enabled in the create isolated network form.
The above check hides it if all the zones have sg enabled or there are no zones present

ravening: Yeah that works fine. I have made the code changes and tested it in setup with two zones. one with sg disabled and 1 with sg enabled . It doesnt show "zone name" with sg enabled for isolated networks. I have added screenshots

ravening:
@davidjumani @utchoang Please review it

It doesnt show zone name which has sg enabled for isolated networks

Screenshot 2020-07-27 at 16 09 06


It shows all zones for guest networks

Screenshot 2020-07-27 at 16 09 51

davidjumani:
@ravening That's right. We filter our zones without SG in the view. Might have to remove that filter in CreateL2NetworkForm. Could you remove that ?

ravening:
@davidjumani made changes to CreateL2NetworkForm so that it can be created in all zones

davidjumani:
@blueorangutan package

PR #558, opened 2020-07-28
Fix bug missing Security Group in Advanced Zone with SG enabled

utchoang:
@davidjumani @rhtyd cc @svenvogel 
* Advanced Zone with SG enabled:
image

* Networks step:
image

* Security Group step:
image

### DIFF
--- a/src/views/compute/DeployVM.vue
+++ b/src/views/compute/DeployVM.vue
@@ -238,6 +238,19 @@
                 </template>
               </a-step>
               <a-step
+                v-if="showSecurityGroupSection"
+                :title="$t('label.security.groups')"
+                :status="zoneSelected ? 'process' : 'wait'">
+                <template slot="description">
+                  <security-group-selection
+                    :zoneId="zoneId"
+                    :value="securitygroupids"
+                    :loading="loading.networks"
+                    :preFillContent="dataPreFill"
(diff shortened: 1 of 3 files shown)


src/views/compute/DeployVM.vue
```diff
@@ -774,6 +780,9 @@ export default {
     },
     networkName () {
       return this.$route.query.name || null
+    },
+    showSecurityGroupSection () {
+      return this.networks.length > 0 && this.zone.securitygroupsenabled && this.hypervisor !== 'KVM'
```
davidjumani, line 785: @utchoang Why is KVM excluded for SG Selection ?

utchoang: @davidjumani 
I refer to https://github.com/apache/cloudstack/blob/master/ui/scripts/instanceWizard.js#L177 then I think KVM does not support it.

davidjumani: Got it! But in https://github.com/apache/cloudstack/blob/master/ui/scripts/ui-custom/instanceWizard.js#L1570 it does support it along with multiple IPs

utchoang: @davidjumani I understood. I'll remove `this.hypervisor !== 'KVM'` condition. Thanks.

davidjumani:
@blueorangutan package

PR #559, opened 2020-07-28
Fix global settings name display issue

ravening:
Global setting name is displayed twice

Screenshot 2020-07-28 at 11 44 07

After the fix

Screenshot 2020-07-28 at 11 44 29

### DIFF
--- a/src/components/view/ListView.vue
+++ b/src/components/view/ListView.vue
@@ -68,7 +68,7 @@
         <os-logo v-if="record.ostypename" :osName="record.ostypename" size="1x" style="margin-right: 5px" />
         <console :resource="record" size="small" style="margin-right: 5px" />
 
-        <span v-if="$route.path.startsWith('/globalsetting')">{{ text }}</span>
+        <span v-if="$route.path.startsWith('/globalsetting')"></span>
         <span v-if="$route.path.startsWith('/alert')">
           <router-link :to="{ path: $route.path + '/' + record.id }" v-if="record.id">{{ $t(text.toLowerCase()) }}</router-link>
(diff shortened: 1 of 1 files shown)


src/components/view/ListView.vue
```diff
@@ -68,7 +68,7 @@
         <os-logo v-if="record.ostypename" :osName="record.ostypename" size="1x" style="margin-right: 5px" />
         <console :resource="record" size="small" style="margin-right: 5px" />
 
-        <span v-if="$route.path.startsWith('/globalsetting')">{{ text }}</span>
+        <span v-if="$route.path.startsWith('/globalsetting')"></span>
         <span v-if="$route.path.startsWith('/alert')">
```
davidjumani, line 72: @ravening would be better to make this a v-else-if and leave the prior line as is since we don't need a link and detail view for each global setting

rhtyd:
@ravening can you address review comments, thanks.

src/components/view/ListView.vue
ravening: @davidjumani can explain more? I didnt understand about the v-else-if part

davidjumani: can change it to
```
         <span v-if="$route.path.startsWith('/globalsetting')">{{ text }}</span>
         <span v-else-if="$route.path.startsWith('/alert')">
```
So that the name is plain text and not a router link to a detail view of the global setting which is pretty much just the name

ravening: @davidjumani the line `<span v-if="$route.path.startsWith('/globalsetting')">{{ text }}</span>` is displaying both plain text and the router link as shown in the screenshot above

ravening: @davidjumani done

davidjumani:
@blueorangutan package

PR #565, opened 2020-07-29
package: update dependencies antd to v1.6.4

rhtyd:
If this breaks master, we can move to 1.1 milestone.


rhtyd:
cc @davidjumani @utchoang thnx
@blueorangutan package

rhtyd:
Bit risky to do it now, let's revisit lib/version change post 1.0/GA

PR #572, opened 2020-08-10
Fixing Edit ACL

### LINKED ISSUE #570: Error when editing firewall rule
I have this rule 
rule

the problem is that when I try to edit that rule to set the protocol to all, I get the erro below.

error

it looks like primate is looking for port number eventhough I am trying the set the protocol to all. this works ok in the legacy UI.

davidjumani:
Fixes #570 

Adding defensive checks for form values and refactoring

### DIFF
--- a/src/views/network/AclListRulesTab.vue
+++ b/src/views/network/AclListRulesTab.vue
@@ -46,7 +46,7 @@
             </div>
             <div class="list__container">
               <div class="list__col">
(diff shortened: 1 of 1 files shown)


davidjumani:
@blueorangutan package

rhtyd:
@gaaray2k can you review/test it (you may use the QA env link ^^) thanks

PR #579, opened 2020-08-13
Minor Fix in Autogen

Pearl1594:
Added check to handle undefined parameter error in AutogenView


Pearl1594:
@blueorangutan package

PR #587, opened 2020-08-17
Allow creating snapshot from VM snapshot

ravening:
In old UI we support creating snapshot from
VM snapshot by selecting one of the volume
belonging to VM. Allow the same in new UI

Navigate to Storage -> VM Snapshots -> Select any one -> Click on camera icon which says create snapshot

Screenshot 2020-08-17 at 18 05 25


rhtyd:
@davidjumani can you review this

rhtyd:
@blueorangutan package

src/config/section/storage.js
```diff
@@ -325,6 +326,16 @@ export default {
...
       searchFilters: ['name', 'domainid', 'account', 'tags'],
       actions: [
+        {
+          api: 'createSnapshotFromVMSnapshot',
+          icon: 'camera',
+          label: 'label.action.create.snapshot.from.vmsnapshot',
+          message: 'message.action.create.snapshot.from.vmsnapshot',
+          dataView: true,
+          popup: true,
+          show: (record) => { return record.state === 'Ready' },
```
davidjumani, line 336: @ravening Could you add a check whether the instance hypervisor is KVM since the legacy UI only supports it for KVM.
https://github.com/apache/cloudstack/blob/master/ui/scripts/storage.js#L3220

ravening: @davidjumani I dont find the key `hypervisor` in `record`. So it always returns undefined when I try to access `record.hypervisor`. Any idea on this?

davidjumani: @ravening The hypervisor can be found via the `virtualmachineid` in the record

rhtyd: @ravening @davidjumani can we make it simpler then? maybe change the list response to pass hypervisor and use it here. Note: 1.0/GA ships with 4.15 - so if it makes it simpler, send a PR against 4.14 branch and return the key as part of vmsnapshot list response?

ravening: @rhtyd if we have to return hypervisor in list snapshot response then we need to call listvirtualmachine api for each snpashot response which might increase response time and also burden on mgt server


As @davidjumani suggested, we have virtulamachineid parameter but we need to call listvirtualmachine api to get the hypervisor type. Im trying to make that work

davidjumani: @ravening That might not work since api calls are async and will return after the `show` method has completed execution. It would be better to alter the API response in the backend to include the vm's hypervisor which could be done quite easily without calling further apis.

ravening: @davidjumani @rhtyd so then I will just check `record.hypervisor === 'KVM'` here and add `hypervisor` field to listVMSnapshot response in backend.

I still need to call list virtual machine to get the hypervisor type

davidjumani: @ravening Yes, add that condition, once the backend has been changed to return the hypervisor as well, we can merge this!
There won't be a need to call another API as the hypervisor field will be added in the backend itself

ravening: @davidjumani I have updated the code here. In the meantime I will work on adding hypervisor type in the backend code

davidjumani: @ravening `record.state === 'Ready' && record.hypervisor === 'KVM'` :smile:

rhtyd:
LGTM, wil merge this as soon as we merge the upstream 4.14 PR https://github.com/apache/cloudstack/pull/4275

rhtyd (approved):

rhtyd:
@ravening is this still WIP, or ready?

ravening:
> @ravening is this still WIP, or ready?

@rhtyd this is ready.. waiting for back end pr to get merged

PR #627, opened 2020-08-27
deployvm : Sending a refresh in case autogen hasn't picked up the new VM

### LINKED ISSUE #591: [BUG] VM state in UI not updated when deployment failure
Primate from PR 550

VM deployment is ongoing:
image

On deployment failure a proper message is shown, but the VM state is not updated to "Error", but instead need to refresh the current view:

image

davidjumani:
Fixes #591

### DIFF
--- a/src/views/compute/DeployVM.vue
+++ b/src/views/compute/DeployVM.vue
@@ -1402,6 +1402,10 @@ export default {
               status: 'progress'
             })
           }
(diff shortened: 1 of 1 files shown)


davidjumani:
@blueorangutan package

andrijapanicsb:
LGTM

PR #628, opened 2020-08-27
deployvm : Setting default min cpunumber and memory to zero

### LINKED ISSUE #620: [BUG] Set default CPU speed and CPU core count to zero (if not autopopulated from ASIS offering)
I've NOT selected the EULA agreement, and when I click the button - there is no error shown (was showing/poping up previously)
(same goes for the password length check - there is a red warning around the password-input-field, but no popup error when you try deploying VM without honouring the needed length)

image

davidjumani:
Fixes #620


davidjumani:
@blueorangutan package

davidjumani:
@rhtyd Good to go!

Issue #604, opened 2020-08-21
[BUG] EDIT TEMPLATE not on pair with the old UI


andrijapanicsb:
Primate:
![image](https://user-images.githubusercontent.com/45762285/90913572-bcea3f80-e3dc-11ea-948f-5b692739f88e.png)

Old UI:
![image](https://user-images.githubusercontent.com/45762285/90913627-cf647900-e3dc-11ea-84aa-37c42a393c04.png) 
and
![image](https://user-images.githubusercontent.com/45762285/90913659-dbe8d180-e3dc-11ea-866c-e8624acfc39f.png)

davidjumani:
@andrijapanicsb The permissions such as featured, public, etc are in the update template sharing button
![Screenshot from 2020-08-24 12-36-28](https://user-images.githubusercontent.com/8244774/91014231-a6bfc780-e606-11ea-8516-65baa40e7b9b.png)
Can edit the tags in the card on the left
![Screenshot from 2020-08-24 12-38-27](https://user-images.githubusercontent.com/8244774/91014324-cd7dfe00-e606-11ea-95e8-18cd8c1e7b65.png)

andrijapanicsb:
Um...right - then a false bug for the missing options (except the confusing sshKeyEnabled) - but, from a user point of view, I find these a bit hiden - but can live with it if everyone else is OK - what I mean - template permission DO define a realy "sharing of template with other accounts/domains", while the button "Update template sharing" manages public/featured/extractable (which I don't find to be a "sharing")

Anyway, parking this with you guys.

andrijapanicsb:
cc @rhtyd please see my comment?

rhtyd:
@davidjumani can you check and remove ssh enabled option if not necessary (does it make any difference)? (from both register and edit template forms)

davidjumani:
@rhtyd It's just a field used to describe the template but does not have any effect in the service layer

rhtyd:
@andrijapanicsb please verify against latest and close. Fixed in https://github.com/apache/cloudstack-primate/pull/634

andrijapanicsb:
LGTM - (the ssh stuff renamed, though I still find the permissions vs sharingf buttons vs. the actual options very confusing)

PR #634, opened 2020-08-28
Bug fixes

davidjumani:
Fixes :
https://github.com/apache/cloudstack-primate/issues/604
Also : Hide cance button while deploying a vm

### DIFF
--- a/src/views/compute/DeployVM.vue
+++ b/src/views/compute/DeployVM.vue
@@ -505,7 +505,7 @@
             </a-steps>
             <div class="card-footer">
               <!-- ToDo extract as component -->
-              <a-button @click="() => this.$router.back()" v-if="!this.loading.deploy">
+              <a-button @click="() => this.$router.back()" :disabled="loading.deploy">
                 {{ this.$t('label.cancel') }}
               </a-button>
               <a-button type="primary" @click="handleSubmit" :loading="loading.deploy">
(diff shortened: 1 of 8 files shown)


davidjumani:
@blueorangutan package

src/views/compute/DeployVM.vue
```diff
@@ -505,7 +505,7 @@
             </a-steps>
             <div class="card-footer">
               <!-- ToDo extract as component -->
-              <a-button @click="() => this.$router.back()" :loading="loading.deploy">
+              <a-button @click="() => this.$router.back()" v-if="!this.loading.deploy">
```
rhtyd, line 508: @davidjumani I think disable would make sense instead of hide?

src/views/compute/DeployVM.vue
```diff
@@ -1409,7 +1409,6 @@ export default {
           this.$router.back()
         }).catch(error => {
           this.$notifyError(error)
-        }).finally(() => {
```
rhtyd, line 1412: Shouldn't we reset in finally? Otherwise it will reset only on error. Or, is the logic that on VM deployment we'll change view so this is not necessary?

davidjumani: On success, it goes back to the previous page. If we keep it in finally, the button will show up for a second before the page redirects

davidjumani:
@blueorangutan package

davidjumani:
@blueorangutan package

PR #638, opened 2020-08-31
Filter out storage pool on which the volume currently exists on

Pearl1594:
Remove the primary store on which the volume is currently present on from the list of possible storage pools to migrate volume to to avoid the following error:
image


Pearl1594:
@blueorangutan package

davidjumani:
@blueorangutan package

davidjumani (approved):
LGTM. Thanks @Pearl1594 
@rhtyd Good to go!

PR #642, opened 2020-08-31
images: fix templates/ISOs all filter

shwstppr:
Screenshot from 2020-08-31 18-35-37

Show **All** filter menu option for admins

PR #654, opened 2020-09-01
template/iso listing User role regression from #642

shwstppr:
Fixes template/ISO listing for User role


shwstppr:
@davidjumani please check

@blueorangutan package

davidjumani (approved):
LGTM

PR #661, opened 2020-09-02
Hiding vlan uri for non admin users

### LINKED ISSUE #659: [BUG] Regular users should not see their network's VLAN
Old UI doesn't show it. I'm wondering if we should also hide it:
image

davidjumani:
Fixes #659


davidjumani:
@blueorangutan package

Issue #630, opened 2020-08-27
[BUG] Batch-deleting VMs doesn't show as completed when done


andrijapanicsb:
Had a few running and a few VMs in Error state (error during deploying them, but they exist as VMs in vSphere)

Choosen to batch delete all of them - confirmed in vCenter that all running VMs were stopped, deleted (with those Error VMs not deleted, I've logged a separate bug in CloudStack for that) - but the Primate kept regreshing and showing those previously Running VMs as Stopping, while there are already stopped and deleted in vSphere:

![image](https://user-images.githubusercontent.com/45762285/91497210-d4826600-e8bd-11ea-8895-3bdaba323aaf.png)

andrijapanicsb:
Hitting the Refresh button in instance view - will show zero VMs (as it should), since they are all deleted.

davidjumani:
A simpler fix would be to refresh the list view every 30 seconds or so that the user has the most up to date info at any time, irrespective of any action performed. Thoughts @rhtyd ?

rhtyd:
@davidjumani it may require some framework level fixes - could it be possible to track the list of asyncjobs somehow and then refresh when all have been completed? (instead of blind poling 30s)

davidjumani:
Yes, but I think it's unnecessary. When refreshing the list view every 30 sec, there's always the latest data available to the user

PR #662, opened 2020-09-02
Refresh data every 30 seconds

davidjumani:
Fixes https://github.com/apache/cloudstack-primate/issues/630

### DIFF
--- a/src/views/AutogenView.vue
+++ b/src/views/AutogenView.vue
@@ -654,6 +654,10 @@ export default {
         }
       }).finally(f => {
         this.loading = false
+        // Refresh the data every 30 sec
+        if (!this.$route.path.startsWith('/exception')) {
+          setTimeout(this.fetchData, 30000)
+        }
       })
     },
     closeAction () {


davidjumani:
@blueorangutan package

davidjumani:
@blueorangutan package

rhtyd:
This is not ideal or fail-safe, what-if it takes more than n * 30s for the batched jobs to finished (where n is the number of resources on which a group action was performed)?

davidjumani:
@rhtyd It refreshes the page every 30 seconds irrespective of a batch job to ensure the user always has the latest details available. It also fixes the batch job refresh since the jobs should complete at some time, and within 30 seconds of them completing, the page refreshes

davidjumani:
@blueorangutan package

rhtyd:
@davidjumani if we've https://github.com/apache/cloudstack-primate/pull/665 can we close this?

davidjumani:
@rhtyd I prefer having this rather than #665 since that is for a specific case (We can use parts of that such as the event when an async job has been completed to do additional stuff that might be needed in the future). This one will ensure the user has the most up to date info at all times. Can extend the refresh interval to 60sec if 30 seems too short

rhtyd:
Can you advise a case where doing infinite refresh makes sense, and where the async event based refresh in the other PR doesn't fix refresh issue?

davidjumani:
It makes sense where multiple users are logged in and make changes, so that the current user gets the most up to date info. Can close this if it is a niche case!

rhtyd:
Closing on your remark David

PR #664, opened 2020-09-03
deployvm: Showing additoinal fields in infocard

davidjumani:
Adds additional fields to the Infocard while creating a vm
- Pod
- Cluster
- Host
- Cpu cores
- Cpu speed
- Memory

Screenshot from 2020-09-03 17-16-10

### DIFF
--- a/src/components/view/InfoCard.vue
+++ b/src/components/view/InfoCard.vue
@@ -32,7 +32,7 @@
               <div v-if="['USER.LOGIN', 'USER.LOGOUT', 'ROUTER.HEALTH.CHECKS', 'FIREWALL.CLOSE', 'ALERT.SERVICE.DOMAINROUTER'].includes(resource.name)">{{ $t(resource.name.toLowerCase()) }}</div>
               <div v-else>
                 <h4 class="name">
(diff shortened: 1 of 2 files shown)


davidjumani:
@blueorangutan package

rhtyd:
@davidjumani I think we did not include those details to limit the height of the infocard - check on smaller screen if we scroll it well?
@utchoang can you review and test this?

davidjumani:
@rhtyd The user can scroll on the Infocard. The reason I added this was so that the user can see the host & compute resources allocated on the side rather than scrolling up to verify the data passed
Added images when in a smaller screen
Screenshot from 2020-09-07 14-08-49
Screenshot from 2020-09-07 14-08-59

src/components/view/InfoCard.vue
```diff
@@ -280,6 +280,19 @@
             </div>
           </div>
         </div>
+        <div class="resource-detail-item" v-if="resource.networks && resource.networks.length > 0">
+          <div class="resource-detail-item__label">Network(s)</div>
```
utchoang, line 284: @davidjumani can you add a translation for this text?

davidjumani:
@blueorangutan package

utchoang (approved):
LGTM! @rhtyd.
Thanks @davidjumani

PR #665, opened 2020-09-03
Adding events for async job completion

### LINKED ISSUE #630: [BUG] Batch-deleting VMs doesn't show as completed when done

davidjumani:
An alternative to refreshing the page every 30 seconds #662 
Fixes #630

### DIFF
--- a/src/utils/plugins.js
+++ b/src/utils/plugins.js
@@ -31,6 +31,7 @@ export const pollJobPlugin = {
        * @param {Function} [successMethod=() => {}]
        * @param {String} [errorMessage=Error]
        * @param {Function} [errorMethod=() => {}]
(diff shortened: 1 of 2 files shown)


davidjumani:
@blueorangutan package

rhtyd:
LGTM, I need to test it before merging though

PR #667, opened 2020-09-06
vApps: Send configuration ID

nvazquez:
Send configuration ID on vApps deployment


andrijapanicsb:
testing in master.

andrijapanicsb:
@nvazquez can you explain (in case I need to understand) what does this one do? thx

PR #672, opened 2020-09-08
FIX - Breadcrumb gone

### LINKED ISSUE #653: [BUG] Click Instance --> Click Volume = breadcrumb gone
clicking on the volume of the instance moves the "breadcrumb" to volumes, so you can't click back on the instance whose volume you are viewing atm (I can use the browser back button, but that's not the same)

image

image

utchoang:
@rhtyd cc @svenvogel 
Fixes #653 
I fixed the error by saving breadcrumb data to `$store` and this is fine. But I'm not sure if it works well compared to the old logic. 
Please check to see if it is okay. Thanks.
image

### DIFF
--- a/src/components/widgets/Breadcrumb.vue
+++ b/src/components/widgets/Breadcrumb.vue
@@ -20,19 +20,22 @@
     <a-breadcrumb-item v-for="(item, index) in breadList" :key="index">
       <router-link
         v-if="item && item.name"
-        :to="{ path: item.path === '' ? '/' : item.path }"
+        :to="{ path: item.fullPath === '' ? '/' : item.fullPath }"
       >
         <a-icon v-if="index == 0" :type="item.meta.icon" style="font-size: 16px" @click="resetToMainView" />
         {{ $t(item.meta.title) }}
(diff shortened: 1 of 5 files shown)


rhtyd:
@davidjumani can you review this?

davidjumani:
@blueorangutan package

davidjumani:
@utchoang It would be better to tweak the breadcrumb only if the user is in a detail view. 
When clicking the zone while listing the instances, the breadcrumb shows /instances/zone-name
Screenshot from 2020-09-09 11-14-23

utchoang:
@davidjumani I think this makes it possible for the user to go back to the previous list by clicking on the breadcrumb instead of using the browser's back button.
@rhytd Do you have any comments?

davidjumani:
@utchoang Agreed, but it gives a wrong idea of the path, like TestZone under Instances. Would be better to add to the breadcrumb if only redirected from within a detail view (or not list view). @rhtyd Thoughts ?

utchoang:
closed because @david closed this issue

PR #682, opened 2020-09-09
Display ACL name instead of ID

### LINKED ISSUE #676: [BUG] ACLs - 2 issues with displaying them
VPC, ACL.

issue1: when go to particular VPC, then to network, then click on the network (to mange it, replace acl, whatever...) - there is no ACL shown (to be able to know which ACL is ALREDY applied on the network) - in the old UI, this is shown when managing a network:
image

instead, the ACL is only visible on network listing  -  issue2:
when we select a VPC and click on networks - there is ACL UUID shown, instead of it's name:
image

ravening:
Fixes #676 

In networks tab under vpc, display acl names instead of id's
Also display acl name in network details page

Screenshot 2020-09-09 at 19 19 21

### DIFF
--- a/src/config/section/network.js
+++ b/src/config/section/network.js
@@ -30,7 +30,7 @@ export default {
       permission: ['listNetworks'],
       resourceType: 'Network',
       columns: ['name', 'state', 'type', 'cidr', 'ip6cidr', 'broadcasturi', 'account', 'zonename'],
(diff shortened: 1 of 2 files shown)


ravening:
Only aclid is returned in list networks api and not its name. So Im not sure how to display acl name in network details page

rhtyd:
@blueorangutan package

davidjumani (approved):
LGTM Thanks @ravening 
Although I think that long term, the acl name should be added in the response to prevent multiple api calls

ravening:
@davidjumani modified the code

davidjumani:
@blueorangutan package

PR #686, opened 2020-09-10
FIX - L2 networks - there is nothing to clean-up or make redundant

### LINKED ISSUE #674: [BUG] L2 networks - there is nothing to clean-up or make redundant
Although this is on pair with the old UI - we should probably remove this (or the whole restart button) for L2 networks, as there is no VR inside, so there is nothing to "cleanup" (recreate VR) nor anything to make redundant (as there is no VR). 

In general, what does it mean to restart an L2 network? (should we remove the button)?

image

utchoang:
Fixes #674 
@rhtyd cc @svenvogel 
Hide restart network button for L2 network.

### DIFF
--- a/src/config/section/network.js
+++ b/src/config/section/network.js
@@ -81,7 +81,8 @@ export default {
           icon: 'sync',
           label: 'label.restart.network',
           dataView: true,
-          args: ['cleanup', 'makeredundant']
+          args: ['cleanup', 'makeredundant'],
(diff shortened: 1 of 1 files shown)


davidjumani:
@blueorangutan package

davidjumani (approved):
LGTM! Thanks @utchoang 
@rhtyd Good to go!

Issue #690, opened 2020-09-10
[BUG] Hide "Public" switch if the global setting prohibits "user" making templates public


andrijapanicsb:
![image](https://user-images.githubusercontent.com/45762285/92698486-bea88280-f34c-11ea-9c33-67361da7c088.png)

It's allowed, with a proper error message - not critical, but we should hide it if the global setting value prohibits regular "user" role to make templates public.

andrijapanicsb:
If fixing this - please do the same on "Register template" and Register ISO forms - and same for the Upload from Local (template/iso)

rhtyd:
@andrijapanicsb please test and close against latest master

andrijapanicsb:
LGTM in all possible forms.

PR #693, opened 2020-09-10
Disable users from creating public templates based on global setting

davidjumani:
Fixes https://github.com/apache/cloudstack-primate/issues/690


davidjumani:
@blueorangutan package

PR #698, opened 2020-09-11
listview: Hiding event checkboxes for users

davidjumani:


davidjumani:
@blueorangutan package

Issue #700, opened 2020-09-11
[BUG] "user" role - reset password form broken


andrijapanicsb:
![image](https://user-images.githubusercontent.com/45762285/92876619-9f702a80-f40a-11ea-8316-9f35dde3b96b.png)

as a regular user role - chaning your own pass (works fine when "admin" reset other user's pass.

rhtyd:
@andrijapanicsb please test and close against latest master

andrijapanicsb:
LGTM

PR #701, opened 2020-09-11
fix change password form for users

Pearl1594:
Fixes: https://github.com/apache/cloudstack-primate/issues/700


Pearl1594:
@blueorangutan package

PR #711, opened 2020-09-15
FIX - Reduce max pagination from 500/page to 200/page

### LINKED ISSUE #708: [BUG] PROPOSAL: Reduce max pagination from 500/page to 200/page
- If you choose to list 500 records per page in i.e. Event, the machine with 6 CPU cores @ 4.0 Ghz becomes almost unusable (Firefox does complain that the page is slowing down the computer
- If you list 500 records (events) per page, then select them all (checkbox) and try to delete, yiou get the URL too long error.

image

In general, I think 500 is simply too much, and 200 works acceptably (slow) with no errors seen so far (i.e. I don't get the URL too long etc)

Should we reduce the max records per page from 500 to 200?
@rhtyd @utchoang @davidjumani

utchoang:
@rhtyd cc @svenvogel 
Fixes #708

### DIFF
--- a/src/views/network/IpAddressesTab.vue
+++ b/src/views/network/IpAddressesTab.vue
@@ -272,7 +272,7 @@ export default {
           errorMethod: () => {
             this.fetchData()
           },
-          loadingMessage: `${this.$t('label.acquiring.ip')}} ${this.$t('label.for')} ${this.resource.name} ${this.$t('label.is.in.progress')}}`,
+          loadingMessage: `${this.$t('label.acquiring.ip')} ${this.$t('label.for')} ${this.resource.name} ${this.$t('label.is.in.progress')}`,
           catchMessage: this.$t('error.fetching.async.job.result')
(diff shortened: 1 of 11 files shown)


davidjumani:
@blueorangutan package

src/views/network/IpAddressesTab.vue
```diff
@@ -272,7 +272,7 @@ export default {
           errorMethod: () => {
             this.fetchData()
           },
-          loadingMessage: `${this.$t('label.acquiring.ip')} ${this.$t('label.for')} ${this.resource.name} ${this.$t('label.is.in.progress')}`,
+          loadingMessage: `${this.$t('label.acquiring.ip')}} ${this.$t('label.for')} ${this.resource.name} ${this.$t('label.is.in.progress')}}`,
```
davidjumani, line 275: @utchoang Could you fix the extra braces ?

PR #717, opened 2020-09-15
zone id appears wrong

Pearl1594:
Code shown instead of zone id when number of zones = 1 in deploy VM form
image


Pearl1594:
@blueorangutan package

rhtyd:
Lgtm

PR #728, opened 2020-09-21
FIX - Adjust the style of the row

### LINKED ISSUE #724: [BUG] minor: center the name of the network for L2 network
image

if possible, center (vertically) L2 networks (other networks have their CIDR shown, 2 rows, centered fine vertically, but L2 network doesn't have the CIDR, so single row, and not centered vertically)

utchoang:
Fixes #724 
image


rhtyd:
LGTM
@utchoang can you fix the conflict?
@davidjumani can you review as well?

PR #729, opened 2020-09-21
Add CIDR next to the network name

### LINKED ISSUE #725: [FEATURE] Add CIDR next to the network name, for deploy-as-is templates
For normal templates, when user chooses some networks to attach the VM to, there is CIDR of that network shown to help user.

With deploy-as-is templates, CIDR is not shown - let's add it when possible - to look something like this:

image

utchoang:
Fixes #725 
@rhtyd cc @svenvogel I completed this task. Please review it. Thanks.
image

### DIFF
--- a/src/views/compute/DeployVM.vue
+++ b/src/views/compute/DeployVM.vue
@@ -277,11 +277,14 @@
                           ]"
                           :placeholder="nic.networkDescription"
                           :filterOption="(input, option) => {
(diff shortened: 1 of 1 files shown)


rhtyd:
@andrijapanicsb @davidjumani can you review this?
@blueorangutan package

davidjumani (approved):
Code LGTM @andrijapanicsb could you verify ?

Issue #753, opened 2020-09-25
[BUG]  Getting 404 after deleting a VM


andrijapanicsb:
VM deletion in progress:

![image](https://user-images.githubusercontent.com/45762285/94292071-c1b28e00-ff5c-11ea-8162-06bed43e3ee0.png)

After it's done, instead of going back to the list of instances, I get 404:

![image](https://user-images.githubusercontent.com/45762285/94292127-d55df480-ff5c-11ea-8e74-057177852431.png)

rhtyd:
@andrijapanicsb @DaanHoogland please test and close against latest master

DaanHoogland:
stillhappens in latest:
<img width="1209" alt="image" src="https://user-images.githubusercontent.com/2486961/94432692-6b7d5f00-0197-11eb-8e88-62693f967f29.png">

<img width="490" alt="image" src="https://user-images.githubusercontent.com/2486961/94432735-7a641180-0197-11eb-98be-287084aef28d.png">

rhtyd:
@utchoang can you check this?

rhtyd:
@andrijapanicsb please test and close

andrijapanicsb:
LGTM, thx

PR #761, opened 2020-09-29
Fix reload data after delete vm

utchoang:
@blueorangutan package

### DIFF
--- a/src/utils/plugins.js
+++ b/src/utils/plugins.js
@@ -66,7 +66,7 @@ export const pollJobPlugin = {
             key: jobId,
             duration: 2
           })
-          eventBus.$emit('async-job-complete')
+          eventBus.$emit('async-job-complete', action)
           successMethod(result)
         } else if (result.jobstatus === 2) {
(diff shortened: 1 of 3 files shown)


utchoang:
@blueorangutan package

utchoang:
Fixes #753 
@rhtyd @davidjumani cc @svenvogel 
Please review it. Thanks.

davidjumani:
@blueorangutan package

davidjumani (approved):
LGTM! Thanks @utchoang

rhtyd:
@davidjumani is this good to go, you've tested it? Or, will you explore a simpler fix?

davidjumani:
@rhtyd I think this is fine, since not calling the pollJob would result in the progress notification not being displayed

davidjumani:
I also think similar changes might be required. After an async job completes the page refreshes. However, it shouldn't if the user is not on the same page. Eg: After deleting a VM, the page should not refresh if the user is viewing a template

PR #768, opened 2020-09-29
Load balancing and Port forwarding are not displayed properly

### LINKED ISSUE #766: [BUG] Not able to configure Load Balancer rules in isolated network
**Describe the bug**
I created an isolated network and I dont see any option to create Load balancer rules.
Only source NAT is enabled in that network

The issue mentioned in https://github.com/apache/cloudstack-primate/issues/678 states that it should be hidden if PF or static nat is enabled. In my case none of them are enabled and still I dont see any options.

This is blocking testing the HAproxy Loadbalancer customization.

**To Reproduce**

1. Log in as a User or an Admin
2. Go to Networks
3. Click on Add networks and create a new Isolated network
4. Select the isolated network created above. Click on public IP address and select the IP address on which Source NAT is enabled
5. I dont see any tab for Load balancing

**Expected behavior**
I should see a tab for load balancing

**Screenshots**
New ui

Screenshot 2020-09-29 at 15 19 00

I do see load balancing option in old ui

Screenshot 2020-09-29 at 15 19 20

Screenshot 2020-09-29 at 15 38 30

ravening:
Fixes #766 

For isolated networks, load balancing, port forwarding and firewall
options are not displayed.
For source nat ip of vpc, all options are displayed but it should not be

Below are the expected actions according to legacy UI

for isolated networks

if static nat is disabled then display all options
if static nat is enabled then display only firewall tab

for VPC networks

for source nat IP, dont display any option
for other ip's firewall is always disabled.
Load balancing and port forwarding options are mutually exclusive

### DIFF
--- a/src/components/view/ResourceView.vue
+++ b/src/components/view/ResourceView.vue
@@ -120,8 +120,17 @@ export default {
         if (this.resource && this.resource.virtualmachineid && this.resource.vpcid) {
           return false
         }
-        if (this.resource && this.resource.vpcid && tab.name !== 'firewall') {
+        // dont display any option for source NAT IP of VPC
+        if (this.resource && this.resource.vpcid && !this.resource.issourcenat && tab.name !== 'firewall') {
           return true
+        }
+        // display LB and PF options for isolated networks if static nat is disabled
+        if (this.resource && !this.resource.vpcid) {
+          if (!this.resource.isstaticnat) {
+            return true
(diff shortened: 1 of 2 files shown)


davidjumani:
@blueorangutan package

davidjumani (approved):
Verified, LGTM. However, don't like the idea of filtering out tabs in multiple views

PR #776, opened 2020-09-30
Fix dedicate zone to a domain / account

Pearl1594:
Fixes zone dedication

When a zone is attempted to be dedicated to a domain, it fails with the following error:
` Unable to execute API command dedicatezone due to missing parameter domainid`

image


Pearl1594:
@blueorangutan package

davidjumani (approved):
Code LGTM

PR #777, opened 2020-09-30
Set deploy-as-is to default on VMware

nvazquez:
Do not set dpeloy-as-is as optional - use by default on VMware

### DIFF
--- a/src/views/image/RegisterOrUploadTemplate.vue
+++ b/src/views/image/RegisterOrUploadTemplate.vue
@@ -198,13 +198,6 @@
             </a-form-item>
           </a-col>
         </a-row>
-        <a-row :gutter="12" v-if="hyperVMWShow && currentForm === 'Create' && deployAsIsSupported">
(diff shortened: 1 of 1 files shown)


rhtyd:
@andrijapanicsb can you review/test this and against expected feature for upstream?

rhtyd:
@blueorangutan package

shwstppr:
@nvazquez @rhtyd will this make Primate UI backward incompatible?

rhtyd:
Yes @shwstppr. Other than checking for backend version there'll be no way to toggle the behaviour. I see many other backend/API changes that were implemented in 4.15 so largely Primate 1.0/GA will work with only 4.15 onwards, which we can speficy in the release notes - http://docs.cloudstack.apache.org/en/latest/installguide/primate.html#requirements

@nvazquez is this ready for merging?

nvazquez:
Yes @rhtyd its good to go

rhtyd (approved):

PR #784, opened 2020-10-02
Confirmation by pressing Space on confirm screens

### LINKED ISSUE #781: [BUG] Confirmation by pressing Space on confirm screens
**Describe the bug**
Some usability from old ui can be implemented here as well. When ever we have a confirmation screen we need to click with the mouse, instead as per the old ui we can press Space to toggle the confirmation of expunge etc.. 

**To Reproduce**
Expunge a volume
Delete a VM 
and more...

**Screenshots**
Screenshot 2020-10-01 at 11 17 06

utchoang:
Fixes #781 
@rhtyd cc @svenvogel


utchoang:
@blueorangutan package

rhtyd:
Good to have, could this cause any regressions?

utchoang:
@rhtyd I think it won't cause a regression. Thanks.

davidjumani:
@utchoang I think what @borisstoyanov meant was that hitting space toggles the value of the first button / checkbox / switch. An autofocus on the first form item would be an appropriate fix

utchoang:
@blueorangutan package

davidjumani (approved):
LGTM. Need to check whether changes might be needed in other custom forms
Thanks @utchoang

PR #799, opened 2020-10-12
Physical network - FIX - VLXAN needs be removed as isolation method from non-Guest traffic type

### LINKED ISSUE #794: [BUG] VLXAN needs be removed as isolation method from non-Guest traffic type
seems that VxLAN is uncodnitionally displayed, even when no network trafic TYPE is selected

AFAIK, VxLAN is only supported as an isolation method for Guest traffic -not for any other kind of traffic

image

utchoang:
Fixes #794 
image
image
image


utchoang:
@blueorangutan package

rhtyd:
Sorry @utchoang the fix was not well explained in the bug ticket. We would want to allow both vlan/vxlan for guest traffic etc. It is possible to have this, also as per API docs:
```
isolationmethod          string   Used if physical network has multiple iso
                                  lation types and traffic type is public.
                                   Choose which isolation method. Valid op
                                  tions currently 'vlan' or 'vxlan', defau
                                  lts to 'vlan'.
```
Due to this I'll need to close this PR.

PR #800, opened 2020-10-12
kubernetesservicetab: Selecting the right ip

davidjumani:
Fixes KubernetesServiceTab from picking up the first ip in the list of all ips
In the listKubernetesCluster response, there is no field associatednetworkid but only network id. KubernetesServiceTab tries to read associatednetworkid instead of networkid which results in all the ip addresses being returned, and in turn selects the first one. This PR fixes it by ensuring the right network is selected and then selects the source nat ip which has all the associated FW, PF and LB rules

### DIFF
--- a/src/views/compute/KubernetesServiceTab.vue
+++ b/src/views/compute/KubernetesServiceTab.vue
@@ -108,13 +108,13 @@
           </template>
         </a-table>
       </a-tab-pane>
-      <a-tab-pane :tab="$t('label.firewall')" key="firewall">
+      <a-tab-pane :tab="$t('label.firewall')" key="firewall" v-if="publicIpAddress">
         <FirewallRules :resource="this.publicIpAddress" :loading="this.networkLoading" />
       </a-tab-pane>
-      <a-tab-pane :tab="$t('label.portforwarding')" key="portforwarding">
+      <a-tab-pane :tab="$t('label.portforwarding')" key="portforwarding" v-if="publicIpAddress">
         <PortForwarding :resource="this.publicIpAddress" :loading="this.networkLoading" />
       </a-tab-pane>
-      <a-tab-pane :tab="$t('label.loadbalancing')" key="loadbalancing">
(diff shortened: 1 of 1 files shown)


davidjumani:
@blueorangutan package

src/views/compute/KubernetesServiceTab.vue
```diff
@@ -327,14 +327,14 @@ export default {
...
-          params.associatednetworkid = this.resource.associatednetworkid
+        if (this.isValidValueForKey(this.resource, 'networkid')) {
+          params.associatednetworkid = this.resource.networkid
         }
       }
       api('listPublicIpAddresses', params).then(json => {
         const ips = json.listpublicipaddressesresponse.publicipaddress
         if (this.arrayHasItems(ips)) {
-          this.publicIpAddress = ips[0]
+          this.publicIpAddress = ips.filter(x => x.issourcenat)[0]
```
rhtyd, line 337: Does this break on shared networks, are shared networks supported for CKS @shwstppr? What about VPCs?

shwstppr: @rhtyd we support shared network. VPC isn't supported at the moment

shwstppr: @davidjumani as Rohit pointed this might break for shared networks as we don't need source NAT for k8s cluster. Were you able to test that?

shwstppr (changes requested):
@davidjumani maybe we can have a fallback for IP when not source NAT. If you have already tested that code LGTM

src/views/compute/KubernetesServiceTab.vue
davidjumani: From what I've seen, it does use the source nat ip of the network
https://github.com/apache/cloudstack/blob/master/plugins/integrations/kubernetes-service/src/main/java/com/cloud/kubernetes/cluster/actionworkers/KubernetesClusterStartWorker.java#L376

davidjumani:
@blueorangutan package

shwstppr (approved):
LGTM

Issue #801, opened 2020-10-12
UI missing - Cannot select option private registry for kubernetes cluster.


tungntvn23:
- Cannot select optional private registry when create kubernetes cluster.

+ New UI
![image](https://user-images.githubusercontent.com/46994732/95721211-75758680-0c9c-11eb-9c7d-9cba214c5512.png)

+ Old UI
![image](https://user-images.githubusercontent.com/46994732/95721270-8aeab080-0c9c-11eb-97df-73daec60fce2.png)

rhtyd:
Was the private registry feature implemented in backend cc @davidjumani @shwstppr @Pearl1594 ?

rhtyd:
Please triage and then fix for 1.0 or move to 1.1 if the backend does not support private registry feature.

rhtyd:
Fixed in #802 

@tungntvn23 can you test latest master and re-open if you're able to reproduce it again? Thanks.

PR #802, opened 2020-10-12
allow adding docker Private registry details while deploying cluster

Pearl1594:
Addresses the following:
1. Allows users to add private registry - if experimental features flag is enabled
2. Sets min number of master nodes to 2 if HA is enabled

Added the Private registry flag:
image

For HA, made the default value of number of master nodes to 2:
Default initial value:
image

Setting it to a value lesser than 2 (which ideally isn't HA), will prompt the user:
image

### DIFF
--- a/src/components/view/ListView.vue
+++ b/src/components/view/ListView.vue
@@ -439,7 +439,9 @@ export default {
         value: this.editableValue
       }).then(json => {
         this.editableValueKey = null
-
+        if (record.name === 'cloud.kubernetes.cluster.experimental.features.enabled') {
+          this.$store.getters.features.kubernetesclusterexperimentalfeaturesenabled = this.editableValue
+        }
         this.$message.success(`${this.$t('message.setting.updated')} ${record.name}`)
         if (json.updateconfigurationresponse &&
           json.updateconfigurationresponse.configuration &&
--- a/src/components/view/ListView.vue
+++ b/src/components/view/ListView.vue
@@ -439,9 +439,7 @@ export default {
         value: this.editableValue
       }).then(json => {
         this.editableValueKey = null
-        if (record.name === 'cloud.kubernetes.cluster.experimental.features.enabled') {
-          this.$store.getters.features.kubernetesclusterexperimentalfeaturesenabled = this.editableValue
-        }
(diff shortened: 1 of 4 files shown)


davidjumani:
Fixes https://github.com/apache/cloudstack-primate/issues/801

Pearl1594:
@blueorangutan package

src/views/compute/CreateKubernetesCluster.vue
```diff
@@ -307,6 +366,7 @@ export default {
   },
   methods: {
     fetchData () {
+      console.log(this.$store.getters.features)
```
shwstppr, line 369: not really needed

shwstppr (approved):
LGTM except for one minor nit

Pearl1594:
@blueorangutan package

src/components/view/ListView.vue
```diff
@@ -439,7 +439,7 @@ export default {
         value: this.editableValue
       }).then(json => {
         this.editableValueKey = null
-
+        this.$store.dispatch('RefreshFeatures')
```
rhtyd, line 442: @Pearl1594 why are we doing this here? Assuming we do list capabilities fetching already on log in?

Pearl1594: @rhtyd If this isn't done, then unless one logs out and logs back in or does a hard refresh, the store isn't going to be updated with the new value of 'kubernetesclusterexperimentalfeaturesenabled' in the capabilities list

src/store/modules/user.js
```diff
@@ -274,6 +274,16 @@ const user = {
           reject(error)
         })
       })
+    },
+    RefreshFeatures ({ commit }) {
+      return new Promise((resolve, reject) => {
+        api('listCapabilities').then(response => {
+          const result = response.listcapabilitiesresponse.capability
+          commit('SET_FEATURES', result)
+        }).catch(error => {
```
davidjumani, line 283: @Pearl1594 Would not resolving the promise cause any issues ? (Promise state, GC, etc)

Pearl1594:
@blueorangutan package

shwstppr (approved):
lgtm

PR #813, opened 2020-10-21
Fix - Incorrect results of list view when refreshing browser

utchoang:
@rhtyd cc @svenvogel 
When URL has query params (page, page size) and if I refresh the browser then listview result will incorrect
image
image


utchoang:
@blueorangutan package

davidjumani (approved):
LGTM Thanks @utchoang

PR #818, opened 2020-10-23
Use scaleVirtualMachine API instead of changeServiceForVirtualMachine

Pearl1594:
Align with Legacy UI and call scaleVirtualMachine API when attempting scaling

### DIFF
--- a/src/views/compute/ScaleVM.vue
+++ b/src/views/compute/ScaleVM.vue
@@ -153,29 +153,22 @@ export default {
     handleSubmit () {
       this.loading = true
 
-      const apiName = this.resource.state === 'Stopped' ? 'changeServiceForVirtualMachine' : 'scaleVirtualMachine'
       if ('cpuspeed' in this.selectedOffering && this.selectedOffering.iscustomized) {
         delete this.params[this.cpuSpeedKey]
       }
 
-      api(apiName, this.params).then(response => {
-        if (apiName === 'scaleVirtualMachine') {
-          const jobId = response.scalevirtualmachineresponse.jobid
-          if (jobId) {
-            this.$pollJob({
-              jobId,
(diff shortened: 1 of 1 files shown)


Pearl1594:
@blueorangutan package

GabrielBrascher:
For reference, here follows some CloudStack issues raised due to changeServiceForVIrtualMachine API command problems:

- https://github.com/apache/cloudstack/issues/3110
- https://github.com/apache/cloudstack/issues/4125

davidjumani:
Thanks @GabrielBrascher There were a few other reasons we went with changeServiceForVirtualMachine which I can't recall atm

rhtyd:
@davidjumani @Pearl1594 @shwstppr can you share reasons for the change - thanks. I see legacy UI uses the scale API, is that it?

shwstppr (approved):
LGTM. Consistent with legacy UI. As per https://github.com/apache/cloudstack/issues/4125 changeServiceForVirtualMachine seems to be having issues and `scaleVirtualMachine` handles both running and stopped VMs.
@davidjumani if there is any corner case failing we can handle in backend

Pearl1594:
@rhtyd 
1. changeServiceForVirtualMachine isn't used by the legacy UI either when we attempt scaling
2. scaleVirtualMachine does what changeServiceForVirtualMachine does - i.e., updates the service offering, and in addition to that, it also performs volume resizing.

GabrielBrascher (approved):
Code LGTM.

CloudStack UI currently also does not use the `changeServiceForVIrtualMachine`. Additionally, there are a couple of open issues regarding `changeServiceForVIrtualMachine` API command problems:
- apache/cloudstack#3110
- apache/cloudstack#4125

PR #839, opened 2020-10-28
showing ostypeid input even if we use vmware because cloudstack api requires it

alexandru-bagu:
When trying to create a VMware template, OS is not selectable and so you get an error from cloudstack api. Based on the old UI, I think rootdiskcontrollertype should also be visible for vmware.

Old UI: https://prnt.sc/v8itpf
New UI: https://prnt.sc/v8iu6w

### DIFF
--- a/src/views/image/RegisterOrUploadTemplate.vue
+++ b/src/views/image/RegisterOrUploadTemplate.vue
@@ -208,14 +208,14 @@
           </a-form-item>
         </a-row>
         <a-row :gutter="12" v-if="hyperKVMShow || hyperVMWShow">
-          <a-col :md="24" :lg="24" v-if="hyperKVMShow">
+          <a-col :md="24" :lg="24">
             <a-form-item :label="$t('label.rootdiskcontrollertype')">
               <a-select
                 v-decorator="['rootDiskControllerType', {
                   initialValue: rootDisk.opts.length > 0 ? 'osdefault' : '',
                   rules: [
                     {
-                      required: true,
+                      required: hyperKVMShow,
                       message: `${this.$t('message.error.select')}`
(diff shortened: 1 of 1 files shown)


rhtyd:
@alexandru-bagu actually the latest Primate syncs with CloudStack 4.15; and with 4.15 we're changing the code to get template from the source OVA than user-provided details.

rhtyd:
cc @nvazquez @PaulAngus

alexandrubagu-bss:
> @alexandru-bagu actually the latest Primate syncs with CloudStack 4.15; and with 4.15 we're changing the code to get template from the source OVA than user-provided details.

So primate targets for cloudstack 4.15+ as far as this API is concerned? Are there any other UI changes that are meant to be used with 4.15 APIs only?

alexandru-bagu:
> @alexandru-bagu actually the latest Primate syncs with CloudStack 4.15; and with 4.15 we're changing the code to get template from the source OVA than user-provided details.

So primate targets for cloudstack 4.15+ as far as this API is concerned? Are there any other UI changes that are meant to be used with 4.15 APIs only?

rhtyd:
@alexandru-bagu there are many features/parts that bring 4.15 UI changes with Primate, so while most of Primate may work with old CloudStack versions; we've had many changes in list API responses, pagination etc for this reason we've now put in restriction in the rpm/deb builds of Primate to require min. v4.15 cloudstack-management pkg; advanced users may use Primate via the archive build; otherwise the minimum version change will make it impossible for users to install cloudstack-primate rpm/deb pkg with 4.14 and older cloudstack-management versions. So, we cannot accept this PR as-is, thanks.

PR #844, opened 2020-11-04
Show display text instead of name

ravening:
While creating networks, for network offerings, show meaningful display text
instead of short names which is not much descriptive

Screenshot 2020-11-04 at 17 07 36

### DIFF
--- a/src/views/network/CreateIsolatedNetworkForm.vue
+++ b/src/views/network/CreateIsolatedNetworkForm.vue
@@ -125,7 +125,7 @@
               :placeholder="this.$t('label.networkofferingid')"
               @change="val => { this.handleNetworkOfferingChange(this.networkOfferings[val]) }">
               <a-select-option v-for="(opt, optIndex) in this.networkOfferings" :key="optIndex">
-                {{ opt.name || opt.description }}
+                {{ opt.displaytext || opt.description }}
               </a-select-option>
             </a-select>
           </a-form-item>
--- a/src/views/network/CreateL2NetworkForm.vue
+++ b/src/views/network/CreateL2NetworkForm.vue
@@ -125,7 +125,7 @@
               :placeholder="this.$t('label.networkofferingid')"
               @change="val => { this.handleNetworkOfferingChange(this.networkOfferings[val]) }">
               <a-select-option v-for="(opt, optIndex) in this.networkOfferings" :key="optIndex">
-                {{ opt.name || opt.description }}
+                {{ opt.displaytext || opt.description }}
               </a-select-option>
             </a-select>
           </a-form-item>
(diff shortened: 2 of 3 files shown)


rhtyd (approved):
LGTM, but we need to check how legacy UI shows the name (name vs displayname vs description)

ravening:
> LGTM, but we need to check how legacy UI shows the name (name vs displayname vs description)

@rhtyd legacy ui displays name but it wont make sense just by looking at the name. `displaytext` gives more info about the offering

GabrielBrascher (approved):
LGTM

**+1** on using display name as a user-friendly identification.

To be honest, sometimes I don't see much value in the _displayname_ field; I think that this is indeed how it should be used.

rhtyd:
@ravening we may need to build concensus then, we largely want Primate 1.0/GA to be backwards compatible wrt UX, so changing how a field is displayed may break several people's env who were used to seeing the name. (To satisfy everyone - one solution could be showing the name and in parenthesis the displayname).

@blueorangutan package

rhtyd:
For now, moved to 1.1 as this is not a blocker or critical issue.

src/views/network/CreateIsolatedNetworkForm.vue
```diff
@@ -125,7 +125,7 @@
               :placeholder="this.$t('label.networkofferingid')"
               @change="val => { this.handleNetworkOfferingChange(this.networkOfferings[val]) }">
               <a-select-option v-for="(opt, optIndex) in this.networkOfferings" :key="optIndex">
-                {{ opt.name || opt.description }}
+                {{ opt.displaytext || opt.description }}
               </a-select-option>
```
rhtyd, line 129: @ravening can this be: `opt.displaytext | opt.name | ...` to cover for the case if/when displaytext is empty?

src/views/network/CreateL2NetworkForm.vue
```diff
@@ -125,7 +125,7 @@
               :placeholder="this.$t('label.networkofferingid')"
               @change="val => { this.handleNetworkOfferingChange(this.networkOfferings[val]) }">
               <a-select-option v-for="(opt, optIndex) in this.networkOfferings" :key="optIndex">
-                {{ opt.name || opt.description }}
+                {{ opt.displaytext || opt.description }}
```
rhtyd, line 128: @ravening can this be: `opt.displaytext | opt.name | ...` to cover for the case if/when displaytext is empty?

src/views/network/CreateSharedNetworkForm.vue
```diff
@@ -284,7 +284,7 @@
               :placeholder="this.$t('label.networkofferingid')"
               @change="val => { this.handleNetworkOfferingChange(this.networkOfferings[val]) }">
               <a-select-option v-for="(opt, optIndex) in this.networkOfferings" :key="optIndex">
-                {{ opt.name || opt.description }}
+                {{ opt.displaytext || opt.description }}
```
rhtyd, line 287: @ravening can this be: `opt.displaytext | opt.name | ...` to cover for the case if/when displaytext is empty?

rhtyd:
cc @davidjumani

PR #845, opened 2020-11-04
Display network type in network overview

ravening:
Display "VPC Tier" as type if an isolated network belongs to VPC
else display its regular type

Screenshot 2020-11-19 at 16 35 16

**Backend changes are in https://github.com/apache/cloudstack/pull/4483**

### DIFF
--- a/src/config/section/network.js
+++ b/src/config/section/network.js
@@ -29,7 +29,16 @@ export default {
       icon: 'apartment',
       permission: ['listNetworks'],
       resourceType: 'Network',
-      columns: ['name', 'state', 'type', 'cidr', 'ip6cidr', 'broadcasturi', 'account', 'zonename'],
+      columns: ['name', 'state',
+        {
+          type: (record) => {
+            if (record.vpcid && record.vpcid !== '') {
+              return 'VPC Tier'
+            }
+            return record.type
+          }
+        },
+        'cidr', 'ip6cidr', 'broadcasturi', 'account', 'zonename'],
       details: ['name', 'id', 'description', 'type', 'traffictype', 'vpcid', 'vlan', 'broadcasturi', 'cidr', 'ip6cidr', 'netmask', 'gateway', 'aclname', 'ispersistent', 'restartrequired', 'reservediprange', 'redundantrouter', 'networkdomain', 'zonename', 'account', 'domain'],
       filters: ['all', 'isolated', 'shared', 'l2'],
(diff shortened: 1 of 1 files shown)


src/config/section/network.js
```diff
@@ -29,7 +29,16 @@ export default {
       icon: 'apartment',
       permission: ['listNetworks'],
       resourceType: 'Network',
-      columns: ['name', 'state', 'type', 'cidr', 'ip6cidr', 'broadcasturi', 'account', 'zonename'],
+      columns: ['name', 'state',
+        {
+          type: (record) => {
+            if (record.vpcid && record.vpcid !== '') {
+              return 'VPC Tier'
```
rhtyd, line 36: This is new, all VPC tiers are isolated networks - should this be changed in backend? I worry if it would confuse users.

ravening: @rhtyd this is just a string displaying that it belongs to vpc..I dont think we should change anything in background.
I have lot of networks then we can use this string to quickly identify which ones belong to vpc

rhtyd: Should we add a VPC column then, keep the type as-is, but by introducing a `vpcid` column it will make it clicky too.

rhtyd:
For now, moved to 1.1 as this is not a blocker or critical issue.

src/config/section/network.js
GabrielBrascher: **+1** on having such information on UI as proposed by @rhtyd.
What about having the VPC name provided instead of just pointing it as a _VPC Tier_? Or do you think that it would be too much for this window?

ravening:
@rhtyd @GabrielBrascher I made changes to display vpcname in a new coloumn with a router link to it.
I also made changes to display the domain name as well

davidjumani (changes requested):
@ravening could you add a backward compatibility check similar to https://github.com/apache/cloudstack-primate/blob/master/src/config/section/infra/secondaryStorages.js#L27 if the backend changes don't get into 4.15 ?

davidjumani:
Backend changes merged. Moving to 1.0

rhtyd (approved):

rhtyd:
@davidjumani merging this as Primate 1.0 has become too much dependent on 4.15 (esp. the VM deployment form) and we should encourage backward compatibility from 4.15 onwards. The backend PR will make into 4.15.
@ravening any separate PR to address David's concern is welcome!

PR #848, opened 2020-11-06
Provide option to force delete the project

ravening:
Provide a cleanup flag so that the project will
be deleted only when there are no resources left
in the project. If users click on delete project
by mistake then everything is deleted.

**NOTE: Backend change is in https://github.com/apache/cloudstack/pull/4453**

Screenshot 2020-11-06 at 14 28 35


GabrielBrascher (approved):
Code LGTM

davidjumani:
@blueorangutan package

davidjumani:
@ravening Could you add a check whether the parameter exists for the API, similar to https://github.com/apache/cloudstack-primate/blob/master/src/config/section/compute.js#L112

ravening:
> @ravening Could you add a check whether the parameter exists for the API, similar to https://github.com/apache/cloudstack-primate/blob/master/src/config/section/compute.js#L112
> It breaks on versions without the cleanup parameter

@davidjumani I can add that but this feature is supported 4.15 onwards only. I dont think there will be other version where this is supported

ravening:
> > @ravening Could you add a check whether the parameter exists for the API, similar to https://github.com/apache/cloudstack-primate/blob/master/src/config/section/compute.js#L112
> > It breaks on versions without the cleanup parameter
> 
> @davidjumani I can add that but this feature is supported 4.15 onwards only. I dont think there will be other version where this is supported

@davidjumani made changes

davidjumani:
@blueorangutan package

davidjumani (approved):
LGTM

Issue #852, opened 2020-11-09
[BUG] ldap add account does not return after selecting AnyDomain filter


DaanHoogland:
**Describe the bug**
<img width="940" alt="image" src="https://user-images.githubusercontent.com/2486961/98532340-dbe0cb00-2281-11eb-87eb-06f4b1bafaa9.png">
selecting AnyDomain as filter does not return even with empty result set.

as opposed to: 
<img width="826" alt="image" src="https://user-images.githubusercontent.com/2486961/98532495-13e80e00-2282-11eb-9abe-14ecc78c9e86.png">
or : 
<img width="911" alt="image" src="https://user-images.githubusercontent.com/2486961/98532596-3548fa00-2282-11eb-8666-54d151810277.png">

it is not clear to me yet, what the problematic circumstances are. some combinations of filter and domain do not return. It seems the an empty result set might be the problem.

**To Reproduce**

1. Log in as an Admin
1. Go to 'Accounts --> Add LDAP Account'
1. select different filter/domain combinations

**Expected behavior**
A result set should be shown within reasonable time.

**Screenshots**
see description

**Desktop (please complete the following information):**
 - OS: [e.g. Windows, Mac, iOS, Android with device/screen details if applicable]

**Additional context**

PR #853, opened 2020-11-09
addldapaccount: Fix defensive check

davidjumani:
Fixes https://github.com/apache/cloudstack-primate/issues/852

### DIFF
--- a/src/views/iam/AddLdapAccount.vue
+++ b/src/views/iam/AddLdapAccount.vue
@@ -283,7 +283,7 @@ export default {
       params.domainid = store.getters.userInfo.domainid
       if (domain) {
         const result = this.listDomains.filter(item => item.name === domain)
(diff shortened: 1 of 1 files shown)


davidjumani:
@blueorangutan package

src/views/iam/AddLdapAccount.vue
```diff
@@ -283,7 +283,7 @@ export default {
       params.domainid = store.getters.userInfo.domainid
       if (domain) {
         const result = this.listDomains.filter(item => item.name === domain)
-        if (result) {
+        if (result.length > 0) {
```
DaanHoogland, line 286: how about result == None?

davidjumani: Can't be none. Either an empty list of not

PR #860, opened 2020-11-12
Enable delete icon on detached volumes

RodrigoDLopez:
When you detach some volume, we cannot see the delete button
image

This PR enable us to see this button on detached volumes
image


RodrigoDLopez:
@blueorangutan package

PR #868, opened 2020-11-20
Display IP address while deleting it

ravening:
While deleting an IP address, the status message displays "Undefined" instead of IP.

Screenshot 2020-11-20 at 16 03 05

Navigate to Networks -> Public IP address -> Select any public ip -> Click delete button on top right

After the fix

Screenshot 2020-11-20 at 16 04 31

### DIFF
--- a/src/views/AutogenView.vue
+++ b/src/views/AutogenView.vue
@@ -992,7 +992,8 @@ export default {
           }
         }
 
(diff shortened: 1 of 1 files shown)


davidjumani:
@blueorangutan package

davidjumani (approved):
LGTM

wido:
Advanced Networking also supports IPv6. Shouldn't that be shown here as well?

rhtyd:
@wido can you test and raise a PR?

PR #879, opened 2020-12-01
Fix showing esxi host label instead of vCentre host

davidjumani:
### Before :
Screenshot from 2020-12-01 17-07-35

### After :
Screenshot from 2020-12-01 17-12-49


davidjumani:
@blueorangutan package

Pearl1594 (approved):

PR #880, opened 2020-12-01
Close snapshot form in case of an error

Pearl1594:
This PR :
- Adds missing label:
image

- Closes the create snapshot form in case of any error:
image

### DIFF
--- a/src/views/storage/TakeSnapshot.vue
+++ b/src/views/storage/TakeSnapshot.vue
@@ -184,6 +184,8 @@ export default {
           this.closeAction()
         }).catch(error => {
           this.$notifyError(error)
+        }).finally(() => {
+          this.actionLoading = false
         })
       })
     },
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -1435,6 +1435,7 @@
 "label.networkrate": "Network Rate (Mb/s)",
 "label.networkread": "Network Read",
 "label.networks": "Networks",
+"label.networkspeed": "Network Speed",
 "label.networktype": "Network Type",
 "label.networkwrite": "Network Write",
 "label.new": "New",


Pearl1594:
@blueorangutan package

src/views/storage/TakeSnapshot.vue
```diff
@@ -181,10 +181,9 @@ export default {
               catchMessage: this.$t('error.fetching.async.job.result')
             })
           }
-          this.closeAction()
         }).catch(error => {
           this.$notifyError(error)
-        })
+        }).finally(() => { this.closeAction() })
```
davidjumani, line 186: The error is that `this.actionLoading` is not reset. Would be better to set it back to false in finally so the user can try creating the snapshot again in case of a failure

Pearl1594: It's caused because of both actionLoading not being set to false and not initiating form closure post completion of the task. And merely, closing the form should do, however, I've reset the actionLoading parameter as well.

Pearl1594:
@blueorangutan package

src/views/storage/TakeSnapshot.vue
davidjumani: The form should remain open in case of an error so the user can try again without having to input all the fields all over again

rhtyd:
@davidjumani @Pearl1594 is this good to go?

davidjumani:
@rhtyd Requested changes from @Pearl1594 so that the form is not closed in case of an error so the user can try again without reentering the data all over again

rhtyd:
Thanks for the update @davidjumani

Pearl1594:
@blueorangutan package

davidjumani:
@blueorangutan package

davidjumani (approved):

rhtyd (approved):

PR #885, opened 2020-12-02
createnetworkform: Displaying api errors

davidjumani:
Shows errors while fetching network offerings. Especially noticeable when there are multiple guest physical networks in a zone but are untagged.
Makes life easier for the user so he knows what's wrong rather than just shown empty dropdowns

Screenshot from 2020-12-03 12-46-32

Screenshot from 2020-12-02 15-04-49

### DIFF
--- a/src/views/network/CreateIsolatedNetworkForm.vue
+++ b/src/views/network/CreateIsolatedNetworkForm.vue
@@ -415,6 +415,8 @@ export default {
       this.selectedNetworkOffering = {}
       api('listNetworkOfferings', params).then(json => {
         this.networkOfferings = json.listnetworkofferingsresponse.networkoffering
+      }).catch(error => {
+        this.$notifyError(error)
       }).finally(() => {
         this.networkOfferingLoading = false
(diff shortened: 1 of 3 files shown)


davidjumani:
@blueorangutan package

rhtyd:
Could this break previous behaviour @davidjumani, does old UI show any exception? Also is this shown only to admins?

davidjumani:
@rhtyd It copies the behaviour of the old UI even for regular users

Screenshot from 2020-12-03 12-42-37
Screenshot from 2020-12-03 12-43-28

rhtyd:
@davidjumani actually the error is not very clear, I'll let @DaanHoogland @PaulAngus advise further on this.

davidjumani:
@rhtyd Changing the error response will require a backend fix. This just copies the old UI functionality so the user knows that an error has occurred trying to create a network (while fetching network offerings)

PaulAngus:
+1, reverting back to the old user experience is definitely better than the new UX.

Ideally, we'd find a way such that the admin can't get CloudStack networking in that state, but that's not one for today.

PR #886, opened 2020-12-03
Disable edit and delete of UI read-only settings

Pearl1594:
Added a setting - cpuNumber to the "user.vm.readonly.ui.details" global setting. Doing so should prevent a user from editing the value for the UI, however this isn't the case. This PR attempts to address this issue.

Legacy UI behavior:
image

It disables the options to edit / delete the particular setting

Post Fix - Primate Behavior:
image


Pearl1594:
@blueorangutan package

davidjumani (approved):

rhtyd (approved):

PR #893, opened 2020-12-08
Dont pass account/domainid for upload sslcert in project view

ravening:
In project view, we can upload ssl certificate either by using
projectid or accountid/domainid combination. Since we are passing
all 3 parameters, we cant upload sslcert in project view

Screenshot 2020-12-07 at 14 08 55

### DIFF
--- a/src/config/section/account.js
+++ b/src/config/section/account.js
@@ -160,15 +160,7 @@ export default {
       label: 'label.add.certificate',
       dataView: true,
       args: ['name', 'certificate', 'privatekey', 'certchain', 'password', 'account', 'domainid'],
-      show: (record) => { return record.state === 'enabled' },
-      mapping: {
-        account: {
-          value: (record) => { return record.name }
-        },
-        domainid: {
-          value: (record) => { return record.domainid }
-        }
-      }
+      show: (record) => { return record.state === 'enabled' }
     },
     {
       api: 'deleteAccount',


davidjumani:
@blueorangutan package

davidjumani:
@ravening This will make the form require the user to enter the account as well as domain which is redundant as he's already navigated to the desired account. Instead, would it be better to exclude passing the projectid if the current route contains "account" over at https://github.com/apache/cloudstack-primate/blob/master/src/utils/request.js#L83 ?

ravening:
> @ravening This will make the form require the user to enter the account as well as domain which is redundant as he's already navigated to the desired account. Instead, would it be better to exclude passing the projectid if the current route contains "account" over at https://github.com/apache/cloudstack-primate/blob/master/src/utils/request.js#L83 ?
> 
> 

@davidjumani we don't need to enter anything here. If we are in project view then project id is passed else account/domain will be taken from backend code

If we pass both account and domain in project view then exception is thrown

davidjumani:
@ravening The account will be taken for the current user but say an admin will have to pass the account and domain while adding a certificate for another account, after having navigated to that account

rhtyd:
Ping @ravening any update on this?

weizhouapache:
@ravening @rhtyd @davidjumani 
can we do like this ?
(1) in project view, ssl cert will be uploaded to project account, no matter what current account is.
otherwise there is no way to upload ssl cert to a project.
(2) in normal view, ssl cert will be uploaded to the current account on web page.

PR #894, opened 2020-12-11
Add options for OOBM action commands

GabrielBrascher:
#### Description

Currently the UI allows to call OOBM actions, however, the ADMIN must fill a texbox with the command which might lead to issues in case the letters are not all capital or the command does not exist.

To help ADMINs, this PR adds an options list with the supported Action commands.

### Screenshot

oobm-action

### How this has been tested

Execute OOBM actions via the UI.

### DIFF
--- a/src/config/section/infra/hosts.js
+++ b/src/config/section/infra/hosts.js
@@ -192,6 +192,9 @@ export default {
       mapping: {
         hostid: {
           value: (record) => { return record.id }
+        },
+        action: {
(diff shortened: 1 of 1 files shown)


GabrielBrascher:
@blueorangutan package

rhtyd (approved):

rhtyd:
If old UI gave options, we can treat it as bug - let's merge - LGTM.

PR #896, opened 2020-12-14
createnetwork: Fixing missing vlan parameter

davidjumani:
Fixes inability to specify vlan for isolated networks

Screenshot from 2020-12-14 11-35-27

Also reduces the issue when the isolated tab pops up seconds later

### DIFF
--- a/src/views/network/CreateNetwork.vue
+++ b/src/views/network/CreateNetwork.vue
@@ -18,15 +18,7 @@
 <template>
   <div class="form-layout">
     <a-tabs defaultActiveKey="1" :animated="false">
-      <a-tab-pane :tab="$t('label.isolated')" key="1" v-if="this.isAdvancedZoneWithoutSGAvailable()">
-        <CreateIsolatedNetworkForm
-          :loading="loading"
-          :resource="resource"
-          @close-action="closeAction"
-          @refresh-data="refreshParent"
-          @refresh="handleRefresh"/>
-      </a-tab-pane>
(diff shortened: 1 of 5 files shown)


davidjumani:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

Pearl1594:
@davidjumani I'm unable to reproduce the issue of not being able to pass the vlan id when chosen isolated n/w's offering has specifyVlan set to true.
However, the fix for tab looks good

src/views/network/CreateNetwork.vue
```diff
@@ -18,24 +18,24 @@
 <template>
   <div class="form-layout">
     <a-tabs defaultActiveKey="1" :animated="false">
-      <a-tab-pane :tab="$t('label.isolated')" key="1" v-if="this.isAdvancedZoneWithoutSGAvailable()">
-        <CreateIsolatedNetworkForm
+      <a-tab-pane :tab="$t('label.l2')" key="1">
```
rhtyd, line 21: @davidjumani can the first network/tab be isolated, then l2 and then shared? Most people would want to see isolated by default; also this is reused in the deploy VM form.

davidjumani: It can, but then l2 form pops open first since the if condition on the isolated form takes time to evaluate, and the reordering of the tabs once evaluated looks messy

shwstppr (approved):
LGTM

PR #897, opened 2020-12-15
Fix label on Zone network setup step

GabrielBrascher:
If a zone holds a storage network then the following label issue happens:

image

After fixing the label:

image


GabrielBrascher:
@blueorangutan package

rhtyd (approved):

PR #908, opened 2020-12-31
FIX - Skip add primary storage if user selected to enable local storage for user and system VMs

### LINKED ISSUE #907: [BUG] UI Does not allow to add a zone without primary shared storage
#### Description:

Via the "legacy UI, it is possible to deploy a Zone that does not contain shared primary storage by enabling local storage for user and system VMs; however, it is mandatory to at least deploy one shared primary storage via the new UI.

#### Expected behavior

Allow continuing with the Zone deployment wizard without shared primary storage if the user has selected to enable local storage for user and system VMs

#### Screenshots

Enabling local storage:
image

Shared primary storage:
image

utchoang:
Fixes #907 
Screenshot_2
Screenshot_3

### DIFF
--- a/src/views/infra/zone/ZoneWizardAddResources.vue
+++ b/src/views/infra/zone/ZoneWizardAddResources.vue
@@ -130,6 +130,8 @@ export default {
     steps () {
       const steps = []
       const hypervisor = this.prefillContent.hypervisor ? this.prefillContent.hypervisor.value : null
+      const localStorageEnabled = this.prefillContent.localstorageenabled.value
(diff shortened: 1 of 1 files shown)


utchoang:
@blueorangutan package

GabrielBrascher (approved):
My first idea was to allow both ways, either deploy shared + local, or just local storage.

But this LGTM as well, simple and it works. If the Admin wants to add primary shared storage he/she can do it via the normal process for adding a new primary storage pool.

Any comments, @rhtyd?

PR #915, opened 2021-01-11
Display all data volumes for vm while destroying

ravening:
If a root admin tries to destroy VM of different
domain then the data disks are not displayed and
hence they wont be destroyed.
Make sure that root admin can see all data disks
of a vm while destroying it

Steps to reproduce:

1. Login as domain admin and create a vm with two data disks
2. Now login as ROOT admin and try to destroy that vm

Expected result:
Root admin should see all two data disks

Actual result:
Root admin doesnt see any data disk

After fix:

Screenshot 2021-01-11 at 11 10 39

### DIFF
--- a/src/views/compute/DestroyVM.vue
+++ b/src/views/compute/DestroyVM.vue
@@ -97,7 +97,8 @@ export default {
       api('listVolumes', {
         virtualMachineId: this.resource.id,
         type: 'DATADISK',
-        details: 'min'
+        details: 'min',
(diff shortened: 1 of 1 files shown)


weizhouapache (approved):
lgtm 
this issue does exist in legacy ui as well.

shwstppr:
@blueorangutan package

utchoang (approved):
Code LGTM! Thanks.

PR #911, opened 2021-01-04
Allow enabling network/vpc offering at creation

Pearl1594:
Backend PR: https://github.com/apache/cloudstack/pull/4564

Additional field in Add Network Offering Form:
image

Additional field in Add VPC offering Form:
image

PR #917, opened 2021-01-14
network/VPC offering - Show field only if paramter is present in the API

Pearl1594:
Addresses comment in : https://github.com/apache/cloudstack-primate/pull/911


Pearl1594:
@blueorangutan package

shwstppr (approved):
LGTM

